PR #8, opened 2017-05-18
feat: now the expected folderPath parameter is an id of folder and not a path

doubleface:
This will adapt the konnectors to the real way of running it by the stack: with an id of folder as parameter.
Another part of this is done in the cozy-konnector-template and then in all the konnectors, one more time :-(

### DIFF
--- a/base_konnector.js
+++ b/base_konnector.js
@@ -41,6 +41,17 @@ module.exports = {
           process.exit(0)
         })
         .then(account => {
+          // get the folder path from the folder id and put it in cozyFields.folderPath
+          return new Promise((resolve, reject) => {
+            cozy.files.statById(cozyFields.folderPath, false)
+            .then(folder => {
+              cozyFields.folderPath = folder.attributes.path
+              resolve(account)
+            })
+            .catch(() => reject('NOT_EXISTING_DIRECTORY'))
+          })
+        })
+        .then(account => {
           const requiredFields = Object.assign({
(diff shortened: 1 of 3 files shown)


base_konnector.js
```diff
@@ -41,6 +41,17 @@ module.exports = {
           process.exit(0)
         })
         .then(account => {
+          // get the folder path from the folder id and put it in cozyFields.folderPath
+          return new Promise((resolve, reject) => {
+            cozy.files.statById(cozyFields.folderPath, false)
```
gregorylegarec: If I understand well, `folderPath` is actually an id ? Is it expensive to rename it ?

doubleface: Yes.. You have to change all the konnectors...

doubleface:
End of sprint merge... Sorry for the reviewers

base_konnector.js
gregorylegarec: So maybe we may consider this issue the other way : why do we have an id instead of a path here ?

doubleface: The back insisted on passing a folder id to avoid exotic path which could be a security problem (like the .cozy_apps path for example.

gregorylegarec: @doubleface Ok thanks !

So I think that we'll have to migrate the `folderPath` field to `folderId` in the future, or at least specify it as `deprecated`.

doubleface: I agree

PR #11, opened 2017-05-22
Adds event model

### LINKED ISSUE #9: Event model missing
Hi, I'm trying to migrate an event konnector (Meetup) but the `event` model is missing from the lib.

I tried to look how those were made, but I'm quickly lost with the `cozyclient` API and didn't find any doc to give me hints.

Is someone already doing this? If not, I could try but some documentation on the client API would be more than welcome 😄 

Thanks

JbIPS:
First attempt at adding the model. I'm still testing it but this will let anyone have a look and maybe spot mistakes 😃 

Closes #9

### DIFF
--- a/models/event.js
+++ b/models/event.js
@@ -1,0 +1,83 @@
+const cozy = require('../cozyclient');
+const DOCTYPE = 'io.cozy.events';
+
+module.exports = {
+  displayName: 'Event',
+  doctype: DOCTYPE,
+  all (callback) {
+    cozy.data.defineIndex(DOCTYPE, [])
+    .then((index) => {
+      return cozy.data.query(index, {});
+    })
+    .then((events) => {
+      callback(null, events);
+    })
+    .catch((err) => {
+      callback(err);
+    });
+  },
+
+  byCalendar (calendarId, callback) {
+    cozy.data.find(DOCTYPE, calendarId)
+    .then((event) => {
+      callback(null, event);
+    })
+    .catch((err) => {
+      callback(err);
+    });
+  },
+
+  getInRange (options, callback) {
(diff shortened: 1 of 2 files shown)


models/event.js
```diff
@@ -0,0 +1,83 @@
+const cozy = require('../cozyclient');
+const DOCTYPE = 'io.cozy.events';
+
+module.exports = {
+  displayName: 'Event',
+  doctype: DOCTYPE,
+  all (callback) {
+    cozy.data.defineIndex(DOCTYPE, [])
```
doubleface: Did you check it works? I think a defineIndex without attribute does not work to get all the elements. You can get an example of implementation [here](https://github.com/JbIPS/cozy-konnector-libs/blob/30b15f71b538e339f0d82fdf76c7a573f48dcf93/models/base_model.js#L9).

And better, you could directly use the baseModel and not have to reimplement everything.
You can use the baseModel like this : 

```javascript
const baseModel = require('./base_model')
module.exports = baseModel.createNew({
  name: DOCTYPE,
  // put your code here
})
```

doubleface:
And @JbIPS, thanks for the work by the way :smile:

models/event.js
JbIPS: I tried it with the fixture and it worked, but I'll try with your implementation 😉

doubleface: The cozy-client-js stub (which uses fixture file) does not use the parameters and then it is not a very good test in this case. This won't work when using the yarn dev command which uses the real implementation of cozy-client-js

doubleface:
I'm sorry, I have to close this PR for now since it is not usable in a connector.

PR #12, opened 2017-06-02
Fix the io.cozy.bank.operations doctype

nono:


y-lohse:
Ping @m4dz because I think he looked at this at some point, not sure if it's of interest.

PR #17, opened 2017-07-07
[WIP] Add a hack to output all debuging information

aenario:
The idea is to output all possible debugging information.

The stack should be responsible for electing what to save or not depending on the instance settings.

This needs to be tested on an actual konnector.


doubleface:
I think this is not needed anymore

PR #18, opened 2017-07-19
CC-23 | Add a way to pass 2FA requisites

m4dz:
We need a way to inform the Cozy-Collect app that a konnector require a 2FA token (e.g. sent by SMS). This PR add a `errors` wrapper that exposes helpers. The first one is `requireTwoFactor`, that logs an `error` message that will be written to the stack's job. Its message is normalized (so Collect can use it to detect that the error is a 2FA requisite). A base64 serialized object can be append to the message string, so the konnector can pass a context (e.g. for _csrf and session token). Collect will then add them to its next request (when the user submit the 2FA token) so the konnector can continue its execution (same session, etc).

It's a little bit messy IMHO, but I can't find a better way to handle this use-case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,15 @@ konnectorLibs.models.file
 You can also find a working example with the Bouygues Telecom konnector: https://github.com/cozy/cozy-konnector-bouyguestelecom.git
 
 
+### About 2FA tokens
+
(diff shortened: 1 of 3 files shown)


nono:
@doubleface, @aenario: ping

PR #26, opened 2017-08-25
Apply recent fixes about optional path + add travis script TO V3 BRANCH

CPatchane:


libs/BaseKonnector.js
```diff
@@ -30,23 +30,28 @@ module.exports = class baseKonnector {
       this.accountId = cozyFields.account
```
CPatchane: Not about my PR here by shouldn't we suppose to get the account ID directly from `account._id` here? It seems less error prone to me to use the last account fetched IMHO

ptbrowne:
Thanks @CPatchane !

PR #27, opened 2017-08-28
Just some cleaning and linting

doubleface:


ptbrowne:
Merging because Travis build is failing because of timezones

PR #28, opened 2017-08-30
Replace defineIndex by findAll in base_model

CPatchane:
Also update cozy-client-js to fix an issue when index creation.


CPatchane:
[Erreur d'index](https://trello.com/c/ksPfZfpZ/583-erreur-dindex)

PR #42, opened 2017-10-18
fix: remove object spread in BaseKonnector

doubleface:
This way, new connector developpers are not forced to use babel.

When we take the time to package cozy-konnector-libs with babel we will be able to use syntaxes not deployed with node 6.

At the moment, we take the risk loose new connectors contributors with one more way to fail.


doubleface:
@ptbrowne I did not see you already did this in #41

ptbrowne:
Oops, my bad also, refactor hidden behind a feature :/

PR #46, opened 2017-10-23
fix: do not call fetch when no function is given the the BaseKonnector constructor

doubleface:


libs/BaseKonnector.js
```diff
@@ -31,17 +31,47 @@ const Secret = require('./Secret')
...
-      this.terminate(err.message || err)
-    })
+    if (typeof fetch === 'function') {
+      this.fetch = fetch.bind(this)
+
+      return this.init()
+      .then(requiredFields => this.fetch(requiredFields))
+      .then(() => log('info', 'The connector has been run'))
+      .then(this.end)
+      .catch(this.fail)
```
ptbrowne: The connector has been run will be printed 2 times here.

ptbrowne: Can we put 
```
      return this.init()
      .then(requiredFields => this.fetch(requiredFields))
      .then(() => log('info', 'The connector has been run'))
      .then(this.end)
      .catch(this.fail)
```

in a separate method `run` ?

ptbrowne: And still call `run` if `fetch` has been provided ?

PR #47, opened 2017-10-24
feat: now saveFiles checks that the mime type corresponds to the extension of the file

doubleface:


doubleface:
[Identifier un connecteur qui dit 'ok' mais renvoie des pdf cassés (ex: Orange, Digiposte, Sosh)](https://trello.com/c/bQc0ISpC/16-identifier-un-connecteur-qui-dit-ok-mais-renvoie-des-pdf-cass%C3%A9s-ex-orange-digiposte-sosh)

doubleface:
Need to have a new version of cozy-konnector-libs

libs/saveFiles.js
```diff
@@ -56,15 +56,18 @@ const downloadEntry = function (entry, folderPath) {
       })
     })
     .then(fileobject => {
+      checkMimeWithPath(fileobject.attributes.mime, fileobject.attributes.name)
+      checkFileSize(fileobject)
+      return fileobject
```
ptbrowne: Why are the checks here and below ? Since the checks return true/false, I think they do nothing here ?

PR #49, opened 2017-10-30
feat: process.exit on terminate() 📝

gregorylegarec:


libs/BaseKonnector.js
```diff
@@ -158,6 +158,7 @@ class baseKonnector {
     // For version of the stack older than 18bcbd5865a46026de6f794f661d83d5d87a3dbf
     log('error', message)
     log('critical', message)
+    process.exit(1)
```
doubleface: I would suggest : `setImmediate(() => process.exit(1))` because of https://nodejs.org/dist/latest-v6.x/docs/api/process.html#process_process_exit_code : 

> It is important to note that calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr.

gregorylegarec: That was precisely what I was aiming, avoid current asynchronous operation to end, for example when calling `this.terminate()` in a promise chain.

If I use `setImmediate()`, the method in my `then()` right after my `catch()` is called, so the process does not terminate immediately.

PR #50, opened 2017-10-31
feat: added updateOrCreate

doubleface:
This tool is need to be able to upgrade the maif connector

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ module.exports = {
   linkBankOperations: require('./libs/linkBankOperations'),
   addData: require('./libs/addData'),
   filterData: require('./libs/filterData'),
+  updateOrCreate: require('./libs/updateOrCreate'),
   request: require('./libs/request'),
   retry: require('bluebird-retry')
 }
--- a/libs/updateOrCreate.js
+++ b/libs/updateOrCreate.js
@@ -1,0 +1,37 @@
+/**
+ * The goal of this function is create or update the given entries according to if they already
+ * exist in the cozy or not
+ *
+ * - `entries` is an array of objects with any attributes :
+ *
+ * - `doctype` (string) is the cozy doctype where the entries should be saved
+ *
+ * - `filters` (array) is the list of attributes in each entry should be used to check if an entry
+ *   is already saved in the cozy
+ *
+ * @module updateOrCreate
+ */
+const bluebird = require('bluebird')
+const log = require('./logger').namespace('updateOrCreate')
+const cozy = require('./cozyclient')
+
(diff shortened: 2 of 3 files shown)


libs/updateOrCreate.js
```diff
@@ -0,0 +1,37 @@
...
+ * - `filters` (array) is the list of attributes in each entry should be used to check if an entry
+ *   is already saved in the cozy
+ *
+ * @module updateOrCreate
+ */
+const bluebird = require('bluebird')
+const log = require('./logger').namespace('updateOrCreate')
+const cozy = require('./cozyclient')
+
+module.exports = (entries = [], doctype, filters = []) => {
```
ptbrowne: I would rename `filters` into `idAttributes` since to me filters would be a function.

libs/updateOrCreate.js
```diff
@@ -0,0 +1,37 @@
...
+  return cozy.data.findAll(doctype)
+    .then(existings => bluebird.mapSeries(entries, entry => {
+      log('debug', entry)
+      // try to find a corresponding existing element
+      const toUpdate = existings.find(doc =>
+        filters.reduce((good, filter) =>
```
gregorylegarec: I feel that `good` is a little bit vague and confusing. I suggest `isMatching` instead. What do you think ?

index.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   linkBankOperations: require('./libs/linkBankOperations'),
   addData: require('./libs/addData'),
   filterData: require('./libs/filterData'),
+  updateOrCreate: require('./libs/updateOrCreate'),
   request: require('./libs/request'),
```
ptbrowne: I find the object oriented approach I took in cozy-konnector-linxo easier to understand. You have documents, they have fields that are used for unicity and then you can forget it, you do not have to call `updateOrCreate` knowing which fields should be used. I would be in favor of porting the object oriented Document class inside cozy-konnector-libs.

I think it would also help to reorganize addData/filterData/updateOrCreate into the same file/class.

libs/updateOrCreate.js
gregorylegarec: Good point, at first I thank it was a function too. Also `mathingAttributes` could to the job (to match my other suggestion below. :)

index.js
ptbrowne: https://gitlab.cozycloud.cc/labs/cozy-konnector-linxo/blob/master/helpers/cozy.js

doubleface: I get you point, this won't be a part of this PR since I use to port the MAIF connector now.

PR #51, opened 2017-11-02
fix: allow to override request default configuration 📝

gregorylegarec:
Explicit mapping was disallowing custom configuration for request, for example it was not possible to set `resolveWithFullResponse` to `true`. As we encapsulate the `request-promise` lib, we should match `request` API.

### DIFF
--- a/libs/request.js
+++ b/libs/request.js
@@ -54,19 +54,12 @@ module.exports = function (options = {}) {
 
   if (options.debug) requestdebug(request)
 
-  const requestOptions = {}
-
-  requestOptions.json = options.json
-  requestOptions.jar = options.jar
-  requestOptions.headers = options.headers
-  requestOptions.followAllRedirects = options.followAllRedirects
-  requestOptions.strictSSL = options.strictSSL
-
   if (options.cheerio) {
-    requestOptions.transform = function (body, response, resolveWithFullResponse) {
-      let result = require('cheerio').load(body)
+    options.transform = function (body, response) {
+      // memory issues https://github.com/cheeriojs/cheerio/issues/830
+      let result = require('cheerio').load(response.body)
 
-      if (resolveWithFullResponse === true) {
+      if (options.resolveWithFullResponse) {
         response.body = result
         result = response
       }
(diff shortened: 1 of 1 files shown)


gregorylegarec:
[Améliorer Numéricable](https://trello.com/c/vRy2A685/90-am%C3%A9liorer-num%C3%A9ricable)

libs/request.js
```diff
@@ -54,38 +54,31 @@ module.exports = function (options = {}) {
...
-  const requestOptions = {}
-
-  requestOptions.json = options.json
-  requestOptions.jar = options.jar
-  requestOptions.headers = options.headers
-  requestOptions.followAllRedirects = options.followAllRedirects
-  requestOptions.strictSSL = options.strictSSL
-
   if (options.cheerio) {
-    requestOptions.transform = function (body, response, resolveWithFullResponse) {
```
doubleface: The problem with this change is that the user cannot change resolveWithFullResponse just for one call to request. Example wich I meet a lot in connectors: 

```javascript
const {request} = require('cozy-konnector-libs')
const rq = request({
  cheerio: true,
  json: false
})

rq('https://test.url')
.then($ => {
  const token = $('.token').val()
  return rq({
    resolveWithFullResponse: true
  })
})
.then(response => {
  // test response.statusCode and get cheerio instance in response.body
})
```

To me, the request function in cozy-konnector-libs was just a `request.defaults` shortcut with better defaults for connectors. But maybe that is not clear.

libs/request.js
```diff
@@ -54,38 +54,31 @@ module.exports = function (options = {}) {
...
-      let result = require('cheerio').load(body)
+    options.transform = function (body, response) {
+      // memory issues https://github.com/cheeriojs/cheerio/issues/830
+      let result = require('cheerio').load(response.body)
```
doubleface: If this solves the memory issue, that is cool.

nono:
@gregorylegarec what should we do on this PR?

libs/request.js
gregorylegarec: Not sure, actually. :/

libs/request.js
gregorylegarec: Yes, actually it confused me a lot. IMHO if we provide an instance of `request` we should encapsulate it in another function or object than one called also `request` or `rq`. Also, I am not sure that set a custom config of `request` for every konnector is a good pattern. I was lost when I looked at the `request-promise` documentation because the konnector lib behaviour did not match the documented API. Maybe it's a litte too "magical" to me.

gregorylegarec:
@nono Too old, conflicting with `master`, and break changing existing behaviour. Let's close it. I have a little doubt about Numéricable konnector needing those changes, but if it's the case we will adapt the code to match other konnectors ones.

libs/request.js
ptbrowne: > To me, the request function in cozy-konnector-libs was just a request.defaults shortcut with better defaults for connectors. But maybe that is not clear.

Indeed it is a bit tricky to have a function named exactly like another one and having a different behavior.

> Maybe it's a litte too "magical" to me.

I agree.

PR #52, opened 2017-11-02
doc: migration from v2 to v3 📚

gregorylegarec:


gregorylegarec:
[A trier Doc de migration de cozy-konnector-libs v2 à cozy-konnector-libs v3](https://trello.com/c/tk6QfULw/102-a-trier-doc-de-migration-de-cozy-konnector-libs-v2-%C3%A0-cozy-konnector-libs-v3)

docs/migrations/v2-to-v3.md
```diff
@@ -0,0 +1,42 @@
...
+
+Oviously, start by upgrading the `cozy-konnector-libs` package.
+
+```bash
+$> yarn upgrade cozy-konnector-libs --save
+```
+
+## Deprecated attributes
+Following konnector's attributes may be removed, as they are not used anymore:
+* `color`
```
doubleface: And `name`, `models`, `category`, `dataType`, `customView`

docs/migrations/v2-to-v3.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Removed models
+The following models has been removed and code using them has to be adapted:
+* Bill
+
+## Removed methods
+The following methods has been removed from codebase:
+* `filterExisting`
+* `linkBankOperation`
+These both ones have been directly encapsulated into `saveBills`
```
doubleface: For connectors not directly related to bills, there are saveFiles`, `filterData`, `addData` and `linkBankOperation` which are called by saveBills internally.

docs/migrations/v2-to-v3.md
```diff
@@ -0,0 +1,42 @@
...
+### No more `request`, now use `request`
+The `request` method is now provided directly by `cozy-konnector-libs`, so the following import
+```js
+const request = require('request')
+```
+have to be replaced by
+```js
+const { request } = require('cozy-konnector-libs')
+```
+`request` is a instance of [`request-promise`](https://github.com/request/request-promise) with a custom configuration (`followAllRedirects` is set to `true`).
```
doubleface: Actually, here, request is a function returning a request-promise instance with "better" defaults for connectors, but I see that the naming is misleading. Maybe requestFactory ?

nono:
Ping @gregorylegarec

docs/migrations/v2-to-v3.md
gregorylegarec: Yes maybe a more explicit name could be a good idea, let's keep this change for later, if you're ok.

doubleface: @gregorylegarec But then, could you replace this sentence with something like

```
`request` is a function returning an instance of [`request-promise`](https://github.com/request/request-promise) with a custom configuration (`followAllRedirects` is set to `true`).
```

PR #53, opened 2017-11-09
feat: added fileStreamProcess and contentType as option for saveFiles

doubleface:
This is needed for the migration of the EDF connector.

### DIFF
--- a/libs/saveFiles.js
+++ b/libs/saveFiles.js
@@ -54,9 +54,9 @@ const downloadEntry = function (entry, options) {
         dirID: folder._id,
         contentType: options.contentType
       }
-      if (options.fileStreamProcess) {
+      if (options.postProcessFile) {
         return filePromise
-        .then(data => options.fileStreamProcess(data))
+        .then(data => options.postProcessFile(data))
         .then(data => cozy.files.create(data, createFileOptions))
       }
 
@@ -130,7 +130,7 @@ module.exports = (entries, fields, options = {}) => {
     folderPath: fields.folderPath,
     timeout: options.timeout,
     postProcess: options.postProcess,
-    fileStreamProcess: options.fileStreamProcess,
+    postProcessFile: options.postProcessFile,
     contentType: options.contentType
   }
   return bluebird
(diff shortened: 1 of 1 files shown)


libs/saveFiles.js
```diff
@@ -38,22 +38,29 @@ const sanitizeEntry = function (entry) {
...
         name: getFileName(entry),
-        dirID: folder._id
-      })
+        dirID: folder._id,
+        contentType: options.contentType
+      }
+      if (options.fileStreamProcess) {
+        return filePromise
+        .then(data => options.fileStreamProcess(data))
+        .then(data => cozy.files.create(data, createFileOptions))
```
ptbrowne: Here instead of returning early and creating a new branch in the code, I think it would be simpler to do

```
if (options.fileStreamProcess) {
  filePromise = filePromise.then(data => options.fileStreamProcess(data))
}
```

to process the file, this way you do not have to repeat cozy.files.create(data, createFileOptions).

libs/saveFiles.js
```diff
@@ -38,22 +38,29 @@ const sanitizeEntry = function (entry) {
...
-      return cozy.files.create(rq(reqOptions), {
+      let filePromise = rq(reqOptions)
+      const createFileOptions = {
         name: getFileName(entry),
-        dirID: folder._id
-      })
+        dirID: folder._id,
+        contentType: options.contentType
+      }
+      if (options.fileStreamProcess) {
```
ptbrowne: I think `options.postProcess` would be easier to understand for the user.

ptbrowne: Ah, `postProcess` already exists, why can't it be used here ?

libs/saveFiles.js
doubleface: That is what I did at first, and it works when we use the fileStreamProcessOption. But in the case we do not use it, the result of the request promise is not a stream and cannot be taken by cozy.files.create. Then I add to change the result of the request promise to a stream and I found it much simpler to create a new branch in the code. I suppose I should add some comments.

libs/saveFiles.js
doubleface: I agree with postProcessFile.

You are right saveEntry does too much

PR #61, opened 2017-12-30
feat: 'requestFactory' alias for 'request' lib

doubleface:
- ```requestFactory``` better represents the feature of this lib

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,19 +1,27 @@
 require('./libs/error')
 
 const requestFactory = require('./libs/request')
+const log = require('./libs/logger').namespace('cozy-konnector-libs')
 
 module.exports = {
   BaseKonnector: require('./libs/BaseKonnector'),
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -11,6 +13,7 @@ module.exports = {
   addData: require('./libs/addData'),
   filterData: require('./libs/filterData'),
   updateOrCreate: require('./libs/updateOrCreate'),
-  request: require('./libs/request'),
+  request: requestFactory, // keep the old alias
```
ptbrowne: I would just add a deprecation warning for us not to rely on this deprecated feature.

```
const deprecate = function (message) {
  return function () {
    console.warn(message)
    return wrapped.apply(this, arguments)
  }
}

request : deprecate(requestFactory, 'Use requestFactory instead of request. It will be removed in cozy-konnector-libs@4')
```

doubleface: Right, here it is

PR #62, opened 2017-12-30
feat: deactivate json header if cheerio is activated

doubleface:
json header can cause some trouble on some vendors and we know we do not want json when doing scrapping

### DIFF
--- a/libs/request.js
+++ b/libs/request.js
@@ -49,6 +49,8 @@ module.exports = function (options = {}) {
 
   options = Object.assign(defaultOptions, options)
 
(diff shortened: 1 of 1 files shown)


libs/request.js
```diff
@@ -49,6 +49,8 @@ module.exports = function (options = {}) {
 
   options = Object.assign(defaultOptions, options)
 
+  if (options.cheerio === true && options.json === undefined) options.json = false
```
gregorylegarec: Isn't `if (options.cheerio && !options.json)` a better test ?

doubleface: You're right

PR #73, opened 2018-01-19
Fix attach file document

ptbrowne:
`fileDocument` was not attached to the entry when the file already existed.

This meant that in the `postProcess` function when saving bills, we did not set the `invoice` correctly.

I also added tests. I created them semi-automatically since the 3 tests do the same thing only with different "arguments".

Spec:

### DIFF
--- a/libs/saveFiles.js
+++ b/libs/saveFiles.js
@@ -71,14 +71,14 @@ const downloadEntry = function (entry, options) {
 
       return cozy.files.create(filePromise, createFileOptions)
     })
-    .then(fileobject => {
+    .then(fileDocument => {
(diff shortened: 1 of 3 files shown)


libs/saveFiles.js
```diff
@@ -103,16 +104,18 @@ const saveEntry = function (entry, options) {
         return cozy.files.trashById(file._id)
         .then(() => Promise.reject(new Error('BAD_DOWNLOADED_FILE')))
       }
+      return file
     })
-    .then(() => true, () => false)
```
kosssi: j'avoue que j'avais bloqué sur cette ligne

PR #75, opened 2018-01-20
Now add reimbursements to debit operation (ex doctor) and bills to the credit operation (CPAM reimbursement) It was reversed and the unit tests could not see that

doubleface:
I also added a unit test which checks the content of operations after the linking

### DIFF
--- a/libs/linkBankOperations.spec.js
+++ b/libs/linkBankOperations.spec.js
@@ -1,4 +1,4 @@
-import linkBankOperations, { Linker } from './linkBankOperations'
+import { Linker } from './linkBankOperations'
 
 jest.mock('./cozyclient')
 const cozyClient = require('./cozyclient')
@@ -18,7 +18,6 @@ describe('linker', () => {
   const bill = { amount: 110, _id: 'b1' }
 
   describe('addBillToOperation', () => {
-
     test('operation witout bills', () => {
       const operation = { _id: 123456 }
 
@@ -57,8 +56,7 @@ describe('linker', () => {
   })
 
   describe('addReimbursementToOperation', () => {
-
-    test('operation witout reimbursements', () => {
+    test('operation without reimbursements', () => {
       const operation = { _id: 123456 }
 
       linker.addReimbursementToOperation(bill, operation, operation)
@@ -69,7 +67,7 @@ describe('linker', () => {
         {
           reimbursements: [{
             amount: 110,
-            billId: "io.cozy.bills:b1",
+            billId: 'io.cozy.bills:b1',
             operationId: 123456
           }]
         }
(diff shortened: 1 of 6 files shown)


libs/linker/billsToOperation/operationsFilters.js
```diff
@@ -7,7 +7,7 @@ const { getIdentifiers, getDateRangeFromBill, getAmountRangeFromBill } = require
 
 // constants
 
-const HEALTH_VENDORS = ['Ameli', 'Harmonie', 'Malakoff Mederic'] // TODO: to import from each konnector
+const HEALTH_VENDORS = ['Ameli', 'Harmonie', 'Malakoff Mederic', 'MGEN'] // TODO: to import from each konnector
```
ptbrowne: This could be in a JSON config file. Not blocking.

package.json
```diff
@@ -51,5 +51,8 @@
     "lint": "standard --fix commands/* helpers/* libs/* index.js",
     "test": "jest",
     "doc": "jsdoc2md --template jsdoc2md/README.hbs libs/*.js helpers/*.js > docs/api.md"
+  },
+  "standard": {
+    "globals": ["jest", "describe", "test", "beforeEach", "expect", "it"]
```
ptbrowne: I would prefer to use the [eslint-config-cozy-app](https://www.npmjs.com/package/eslint-config-cozy-app) and prettier since we use it in other projects. Definitevely not blocking.

libs/linkBankOperations.spec.js
```diff
@@ -121,12 +122,20 @@ describe('linker', () => {
...
+      minAmountDelta: 1,
+      maxAmountDelta: 1,
+      pastWindow: 1,
+      futureWindow: 1
+    }
+
+    function updateOperation (doctype, id, attributes) {
+      const operation = operations.find(operation => operation._id === id)
+      Object.assign(operation, attributes)
+      return Promise.resolve(operation)
```
ptbrowne: Since here we mutate objects that could be used for other tests, I would add a beforeEach to reset operations for every test.

libs/linkBankOperations.spec.js
```diff
@@ -136,15 +145,21 @@ describe('linker', () => {
...
         }
       ]
       const options = { ...defaultOptions, identifiers: ['CPAM'] }
       return linker.linkBillsToOperations(healthBills, options)
       .then(result => {
         expect(result).toEqual({
-          b1: { creditOperation: operations[1], debitOperation: operations[0] },
+          b1: { creditOperation: operations[1], debitOperation: operations[0] }
         })
+        expect(operations[0]).toMatchObject({reimbursements: [{
```
ptbrowne: Seems the linter has not gone here.

ptbrowne:
Looking good. I have some remarks and comments but overall LGTM.

PR #87, opened 2018-02-07
feat: add synchronization documentation 📚

gregorylegarec:


gregorylegarec:
[[2j] (3) ETD j'ai exploré la possibilité pour un connecteurs d'avoir un système me permettant de gérer l'état de synchro/historique](https://trello.com/c/r8rpNsCr/845-2j-3-etd-jai-explor%C3%A9-la-possibilit%C3%A9-pour-un-connecteurs-davoir-un-syst%C3%A8me-me-permettant-de-g%C3%A9rer-l%C3%A9tat-de-synchro-historique)

docs/konnectors-history.md
```diff
@@ -0,0 +1,82 @@
...
+  "metadata": {
+    "sync": {
+      "id": "7ee401e841c94159addb47f190903139"
+    }
+  }
+}
+```
+
+The information to save must be determinated by konnector developers. By the way we recommend to use following information, as a basis:
+
```
ptbrowne: s/determinated/determined/

docs/konnectors-history.md
```diff
@@ -0,0 +1,82 @@
...
+| konnector | The slug of the konnector (Example: `trainline`, `freemobile`, `cic`). This could be very useful to retrieve data synchronized with this konnector.
+| last_sync_date | Depending on how a konnector
+
+This is just a suggestion, contributors are let free to determine what data is useful for their own konnector history.
+
+### Helper
+
+#### Saving
+
+As provider of `cozy-konnector-libs`, we should provide a standard way to deal with history, for example with helper dedicated to manage this kind of data.
```
ptbrowne: `with a helper`

ptbrowne:
Thanks for starting the work on that.

It seems unclear in this RFC the problems that are going to solve. I think it would be useful if the introduction part could be expanded with the real world needs.

I guess the questions we are tring to solve here are : 

- How to identify the provider object related to our document  ?
- Is a provider document already in our database ?
- When a document has been updated from the konnector ?

Have I understood well ?

I think the word `history` is a bit confusing :/ When I read history I am thinking of the previous versions of the document but this does not seem to do that. I think if the role of KonnectorHistory is to save metadata, why not simply call it Metadata ?

I think we should also add the state of art and what is thought as shortcomings, for example I know that in the LinxoKonnector we identify if a document has already been in the database via its `linxoId` (related code: https://gitlab.cozycloud.cc/labs/cozy-konnector-linxo/blob/master/helpers/cozy.js#L82). It is declared as an `idAttribute` of the document as it is a technique adopted by the Backbone project.

ptbrowne:
@gregorylegarec I think it would be better until this document is finished to collaborate on Paper, it makes it very easy to add comments on section and to suggest modifications. What do you think ?

docs/konnectors-history.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+The information to save must be determinated by konnector developers. By the way we recommend to use following information, as a basis:
+
+| field | role |
+|-------|------|
+| id        | The id of the document, but given by the external service. If the external service does not provide any id or uuid, it could be interesting to generate one, with an hash of the file for example.
+| konnector | The slug of the konnector (Example: `trainline`, `freemobile`, `cic`). This could be very useful to retrieve data synchronized with this konnector.
+| last_sync_date | Depending on how a konnector
```
CPatchane: I think the sentence is not complete here

gregorylegarec:
@ptbrowne Thanks for your feedback.

> It seems unclear in this RFC the problems that are going to be solved. I think it would be useful if the introduction part could be expanded with the real world needs.

Provide a way to unify how storing information about already synchronized document is done.

> I guess the questions we are tring to solve here are : 
- How to identify the provider object related to our document  ?

IMO, not really. This is an information we shoud store actually, but I don't see any complexity here. The only thing I see is using the slug as reference, what happen when it changes ? Again, we're going back to the id of setting `uuid` for konnectors.

> - Is a provider document already in our database ?

That it. So, how every konnector should deal to "tag" it, and what tool we are giving to contributor to easily retrieve this information.

> - When a document has been updated from the konnector ?

This is a very valuable information, but not mandatory. In my mind it is a konnector concern (every konnector does not need to deal with the last_modification date. By the way, it could be very easy and cheap to always add some date information.

> Have I understood well ?

Maybe you're going a little bit too far, what I have in mind is to propose a really permissive way to store synchronization data, and that every konnector store this data at the same place in documents and with a minimum set of data.

> I think the word `history` is a bit confusing :/ When I read history I am thinking of the previous versions of the document but this does not seem to do that. I think if the role of KonnectorHistory is to save metadata, why not simply call it Metadata ?

I totally agree with you about history. Fist term was "cache" but it was absolutely wrong. History is not exactly what we are doing. By the way, metadata is too vague and can define a lot more than just synchronisation data. That why I'm pretty sure that the attribute `metadata.sync` is the correct place. So we should define the concept as "Synchronisation data" or "sync data". What do you think ?

> I think we should also add the state of art and what is thought as shortcomings, for example I know that in the LinxoKonnector we identify if a document has already been in the database via its `linxoId` (related code: https://gitlab.cozycloud.cc/labs/cozy-konnector-linxo/blob/master/helpers/cozy.js#L82). It is declared as an `idAttribute` of the document as it is a technique adopted by the Backbone project.

I disagree with this technique, first because no attribute should be named with a `attribute` suffix, even if the guys at Backbone are doing this. Second because I am pretty sure that storing all synchronisation data in the same place (and not at the document "root") is the right solution. By the way, as every konnector developer may actually do what he want, Linxo konnectors may keep this pattern as long as we want, or may maintain legacy if we decide to move where the linxoId is stored.

gregorylegarec:
@CPatchane The way of how a konnector is matching already synced data is totally let to the konnector, because we cannot guess how every konnector will work. We could maybe provide a naive matching function which is looking on ids, but I think that we should now only provide a common way to store sync data and to retrieve it. The rest is the konnector concern. At least, for now.

gregorylegarec:
@ptbrowne Totally agree for a paper !

gregorylegarec:
Thanks for your previous feedback @ptbrowne, I rewrote the whole document and proposed and API and a first draft of implementation.

gregorylegarec:
Also, feel free to have a look @doubleface.

docs/synchronization.md
```diff
@@ -0,0 +1,220 @@
...
+      "id": "<entry_id>",
+      "last_sync": "<now>",
+      "fileName": "<fileName>"
+    }
+  }
+}
+```
+
+## Remaining questions
+Do we need to pass as second parameter to `shouldSave`, and `shouldDelete` the whole existing document or just the existing synchronzation data ?
```
CPatchane: IMHO since all sync data should be unique, I think the sync data must be enough. Except for the legacy doc which will need the entire document.

docs/synchronization.md
```diff
@@ -0,0 +1,220 @@
...
+A konnector should provide its own method:
+```js
+{
+  getSyncData: (entry) => ({
+    filename: entry.name
+  })
+}
+```
+
+### idAttribute
```
CPatchane: idAttributes? Since we can have several ones. What about `idEntries`? It's more meaningful for the case here to my mind.

gregorylegarec: As we are already dealing with entities named `entries` in the konnector libs, I think it could be confusing. At first look I was not convinced with `idAttribute` either, because I did not understant well what its meaning was. It means _unique identifier attribute_ and specify the name of the entry attribute which should be use as unique identifier.

docs/synchronization.md
gregorylegarec: You're right about the legacy document.
Also, if we pass only sync data, we could not be able to detect a difference between a non existing document and an existing document without sync data.

ptbrowne:
Thanks for the rewriting @gregorylegarec , it makes much more sense to me !

I think we should have the current options from saveFiles 

* findLegacyDocument
* idAttributes
* shouldSave
* shouldUpdate

as part of a Document class. It would make more sense to me and would allow other places to get those options if need be. It is for me clear that they are method to the document since every one of those methods takes the `entry` in input.

I think that the Model way of organizing code is known to most developers and would help newcomers/contributors to contribute more easily.

```
class Bill extends Document {
  shouldSave: () => {
    return Promise.resolve(moment(this.creationDate).year().isAfter(2010)) 
  },
  shouldUpdate: existingDocument => {
    const hasBeenModified = moment(this.modificationDate).isAfter(existingDocument.metadata.sync.last_sync)
    return Promise.resolve(hasBeenModified)
  },
  idAttribute: 'uuid'
}
```

ptbrowne:
Also, I think that the sync data should be moved one level up and live in the metadata attribute :)

gregorylegarec:
@ptbrowne Thanks, I did not dive deeply into the internal implementation, my thoughts were to add the default method into the `saveFiles` script, but adding default method on Document object seems also pretty consistent.

About inheritance over multiple type of Models, you know my point, I am not a big fan. But the future will tell us what our needs will be.

And yes, sync data is definitely stored into the `metadata` attribute, see the whole example here https://github.com/cozy/cozy-konnector-libs/pull/87/files#diff-4ae1debb5106295eefaa83ad46182733R209.

If you're ok with the main ideas of this documentation, I think we can merge if you agree.

gregorylegarec:
@doubleface Should be nice to have your feedback ! <3

docs/synchronization.md
```diff
@@ -0,0 +1,220 @@
...
+```js
+saveFiles(files, {idAttribute: (entry) => hash(entry)})
+// hash() is a custom method  hashing the current file.
+```
+
+### konnector
+
+The konnector attribute is used as a key reference to retrieve all document synchronized by this konnector.
+
+For now, the `konnector` option will be needed for every call. But it could be interesting to initialize some kind of runner with it, which may encapsulate all the `cozy-konnector-libs` functions.
```
doubleface: I think we can get the slug directly from the account (account_type) and feed saveFiles with it somehow.

docs/synchronization.md
```diff
@@ -0,0 +1,220 @@
...
+```
+When calling this function, it is assumed that `cozy-konnector-libs` first queries the database for an exisring document, based on synchronization data and pass the document resulting the query to the `shouldSave` function, event if this document is `null`.
+
+As a third parameter, a cozyClient instance is passed, it may be used by a konnector to perform another query.
+
+Because of this third parameter, the whole `shouldSave` method is asynchronous.
+
+### shouldUpdate
+
+The `shouldUpdate` function acts exactly like `shouldSave`, except that its used to determine if a document should be updated. By default, we make the choice to never update a document (except if it does not have synchronization data, see `findLegacyDocument` above). But some konnectors like `Linxo` related ones need to update existing documents.
```
doubleface: But the linxo connector does not save files.

docs/synchronization.md
doubleface: I would anyway prefer to pass all the document to let the connector developpers handle some cases we did not think of.

docs/synchronization.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Saving document based on synchronzation options
+
+cozy-konnector-libs must provide some way to easily and transparently save synchronization data. The idea is to provide additional options to the method [`saveEntry`](https://github.com/cozy/cozy-konnector-libs/blob/master/libs/saveFiles.js#L87), and by encapsulation to the method [`saveFiles`](https://github.com/cozy/cozy-konnector-libs/blob/master/libs/saveFiles.js#L149).
+
+These options are:
+
+| Option | Role |
+|--------|---------|
+| disableSyncData | (boolean) Disable storage of  synchronization data. Default: `false` |
```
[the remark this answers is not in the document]
gregorylegarec: Not yet, I just feel someone somehow will need it, but we can totally wait for an explicit request.

docs/synchronization.md
gregorylegarec: `account_type` is supposed to be removed, for now it is just maintained for legacy, as every information is now located in the trigger. As the sync data need the konnector slug, it feels natural to me to pass it as a parameter somewhere. What do you think ?

docs/synchronization.md
gregorylegarec: Well spotted. I don't know how bank operations are saved, could you link me the part in konnector-libs or in the Linxo konnector please ?

docs/synchronization.md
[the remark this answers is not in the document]
gregorylegarec: @doubleface Actually, I think we may also need this option to disable by default the feature when doing first tries and integrations. We could activate it for test on given konnectors.

When the feature will be ready, we will set the default value for `disableSyncData` to `false`.  What do you think ?

docs/synchronization.md
gregorylegarec: I checked how Linxo works, and it is not incompatible, as a document here may have any doctype.

By the way, I am adding that we have to handle synchronisation options not only in `saveFiles` method but also in `addData`.

PR #89, opened 2018-02-15
fix: default log level to debug

### LINKED ISSUE #88: log('debug', 'useful message') does not works on cozy instance in debug mode
After passing a cozy instance to debug mode, by running the following command:

```shell
cozy-stack instances debug <domain> true
```

The instruction `log('debug', 'useful message') does not output anything.

Expected behavior:

debug message is shown

doubleface:
Since the stack already handle logs filtering itself.
Fixes #88

### DIFF
--- a/libs/log-filters.js
+++ b/libs/log-filters.js
@@ -1,11 +1,11 @@
 const levels = {
-  debug: 0,
   secret: 0,
-  info: 10,
-  warn: 20,
-  error: 30,
-  ok: 40,
-  critical: 40
+  debug: 10,
+  info: 20,
+  warn: 30,
+  error: 40,
+  ok: 50,
+  critical: 50
 }
 
 const Secret = require('./Secret')
--- a/libs/logger.js
+++ b/libs/logger.js
@@ -2,10 +2,10 @@
 const { env2formats } = require('./log-formats')
 const { filterLevel, filterSecrets } = require('./log-filters')
 
-const { DEBUG, NODE_ENV, LOG_LEVEL } = process.env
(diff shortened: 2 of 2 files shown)


libs/log-filters.js
```diff
@@ -1,6 +1,6 @@
 const levels = {
-  debug: 0,
   secret: 0,
+  debug: 10,
```
ptbrowne: shouldn't other levels be increased by 10 ?

ptbrowne:
small question otherwise lgtm :)

libs/log-filters.js
doubleface: Right

libs/logger.js
```diff
@@ -2,10 +2,10 @@
 const { env2formats } = require('./log-formats')
 const { filterLevel, filterSecrets } = require('./log-filters')
 
-const { DEBUG, NODE_ENV, LOG_LEVEL } = process.env
+const { NODE_ENV, LOG_LEVEL } = process.env
 const env = (env2formats[NODE_ENV] && NODE_ENV) || 'production'
 
-let level = LOG_LEVEL || (DEBUG && DEBUG.length ? 'debug' : 'info')
+let level = LOG_LEVEL || 'debug'
```
gregorylegarec: Does that mean that by default, every konnector is at `debug` level, and the log filtering is done by the stack ?

doubleface:
That's it

libs/logger.js
doubleface: That's it

PR #91, opened 2018-02-19
Better doc reference

doubleface:
Context: the goal here is to have more content about each feature in the API, with more examples, to make it more like a reference to the connector contributors.

### DIFF
--- a/libs/cozyclient.js
+++ b/libs/cozyclient.js
@@ -1,6 +1,18 @@
 /**
  * This is a [cozy-client-js](https://cozy.github.io/cozy-client-js/) instance already initialized and ready to use
- * @module  cozy-client
+ *
+ * If you want to access cozy-client-js directly, this method gives you directly an instance of it,
+ * initialized according to `COZY_URL` and `COZY_CREDENTIALS` environment variable given by cozy-stack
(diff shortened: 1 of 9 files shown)


libs/cozyclient.js
```diff
@@ -1,6 +1,18 @@
...
- * @module  cozy-client
+ *
+ * If you want to access cozy-client-js directly, this method gives you directly an instance of it,
+ * initialized according to `COZY_URL` and `COZY_CREDENTIALS` environment variable given by cozy-stack
+ * You can refer to the [cozy-client-js documentation](https://cozy.github.io/cozy-client-js/) for more information.
+ *
+ * Example :
+ *
+ * ```javascript
+ * const {clientClient} = require('cozy-konnector-libs')
```
ptbrowne: cozyClient

libs/hydrateAndFilter.js
```diff
@@ -1,21 +1,44 @@
 /**
- * Used not to duplicate data.
+ * This function filters the passed array from data already present in the cozy so that there is
+ * not displicated data in the cozy.
```
ptbrowne: duplicated

PR #92, opened 2018-02-20
Should save, update in hydrateAndFilter

ptbrowne:
Add support for `shouldSave` and `shouldUpdate` in `hydrateAndFilter`.

As a connector developer, you can now choose if an entry needs to be saved/updated.

```js
hydrateAndFilter(entries, doctype, { shouldSave: entry => entry.amount > 100 })
```

`shouldSave` and `shouldUpdate` can also live directly in the object (or better, in
its `prototype`) which allows for `class`ic style : 

```js
class Simpson extends Document {
  shouldSave () => this.amount > 100
}
```

This follows from the ideas from the [RFC for synchronization](https://github.com/cozy/cozy-konnector-libs/pull/87).

It is important for https://trello.com/c/0vcmkkmf/543-2-liaison-intelligente-des-bills-de-sant%C3%A9-avec-un-cr%C3%A9dit-qui-serait-fait-en-une-fois-en-fin-de-mois as we are updating already existing bills with a new attribute. Before, updates were ignored.

### DIFF
--- a/libs/addData.js
+++ b/libs/addData.js
@@ -4,18 +4,19 @@
  * @module addData
  */
 const bluebird = require('bluebird')
+const omit = require('lodash/omit')
 const log = require('./logger').namespace('addData')
 
 module.exports = (entries, doctype) => {
   const cozy = require('./cozyclient')
-  return bluebird.mapSeries(entries, entry => {
+  return bluebird.mapSeries(entries, async entry => {
     log('debug', entry, 'Adding this entry')
-    return cozy.data.create(doctype, entry)
-    .then(dbEntry => {
-      // also update the original entry _id to allow saveBills' linkBankOperation entries to have
-      // an id
-      entry._id = dbEntry._id
-      return dbEntry
-    })
+    const dbEntry = await (entry._id
+      ? cozy.data.update(doctype, entry, omit(entry, '_rev'))
+      : cozy.data.create(doctype, entry))
+    // Also update the original entry _id to allow saveBills'
+    // linkBankOperation entries to have an id
+    entry._id = dbEntry._id
+    return dbEntry
   })
 }
--- a/libs/hydrateAndFilter.js
+++ b/libs/hydrateAndFilter.js
@@ -13,6 +13,7 @@
 
 const bluebird = require('bluebird')
 const log = require('./logger').namespace('hydrateAndFilter')
+const get = require('lodash/get')
 const uniqBy = require('lodash/uniqBy')
 
 /**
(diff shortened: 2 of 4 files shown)


libs/addData.js
```diff
@@ -4,18 +4,19 @@
...
     log('debug', entry, 'Adding this entry')
-    return cozy.data.create(doctype, entry)
-    .then(dbEntry => {
-      // also update the original entry _id to allow saveBills' linkBankOperation entries to have
-      // an id
-      entry._id = dbEntry._id
-      return dbEntry
-    })
+    const dbEntry = await (entry._id
+      ? cozy.data.update(doctype, entry, omit(entry, '_rev'))
```
doubleface: If this function now does some updates, its name does not fit. Plus, we also have a createOrUpdate function.

libs/hydrateAndFilter.js
```diff
@@ -1,45 +1,63 @@
...
-  //  - keys : this is the list of keys used to check that two items are the same
   const keys = options.keys ? options.keys : ['_id']
   const store = {}
 
   log('debug', keys, 'keys')
 
   const createHash = item => {
     return keys.map(key => {
-      let result = item[key]
+      let result = get(item, key)
```
doubleface: Why is lodash/get needed here ? I find it less readable like this.

libs/hydrateAndFilter.js
```diff
@@ -72,29 +90,41 @@ const hydrateAndFilter = (entries, doctype, options = {}) => {
...
-    })
+  const defaultShouldSave = () => true
+  const defaultShouldUpdate = existing => false
+
+  const filterEntries = store => async () => {
+    // Filter out items according to shouldSave / shouldUpdate.
+    // Both can be passed as option or can be part of the entry.
+    return uniqBy(await bluebird.filter(entries, entry => {
+      const shouldSave = entry.shouldSave || options.shouldSave || defaultShouldSave
+      const shouldUpdate = entry.shouldUpdate || options.shouldUpdate || defaultShouldUpdate
```
doubleface: I find it weird that the filter function needs to know if an entry needs to be updated or created and this, again, feels  redundant with the createOrUpdate function. But maybe this is createOrUpdate which is obsolete.
But I this has more meaning in the `Model` vision we can see in your unit tests.

Why not define a `customFilter` option which takes a function and let the connector developer define it's own filter ? And maybe our own `Document` implementation would define these custom filters?

libs/hydrateAndFilter.js
ptbrowne: Because you may want to use a path instead of the name of an attribute.

`keys = ["metadata.version", "name"]` would work here.

libs/addData.js
ptbrowne: Where is this function ?

https://github.com/cozy/cozy-konnector-libs/search?utf8=%E2%9C%93&q=createOrUpdate&type=

What do you think if we rename this function to "createOrUpdateEntries" ?

PR #94, opened 2018-02-20
Handle group amount

ptbrowne:
Allow credit bill to be matched against `groupAmount` attribute of banking operation.

A credit bill has a `groupAmount` attribute in case of bundled reimbursements. This is the attribute that needs to be matched with the banking operation.

### DIFF
--- a/libs/linkBankOperations.spec.js
+++ b/libs/linkBankOperations.spec.js
@@ -169,7 +169,7 @@ describe('linker', () => {
       })
     })
 
-    test('health bills with not debit operation found should associate credit operation', () => {
+    test('health bills with not debit operation found should be associated with a credit operation', () => {
       const healthBills = [
         {
           _id: 'b1',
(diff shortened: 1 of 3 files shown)


libs/linkBankOperations.spec.js
```diff
@@ -192,6 +194,70 @@ describe('linker', () => {
...
+                b2: { creditOperation: operations[1], debitOperation  }
+              })
+              expect(operations[1]).toMatchObject({bills: ['io.cozy.bills:b1', 'io.cozy.bills:b2']})
+              expect(result.b1.debitOperation).toBe(result.b2.debitOperation)
+              expect(result.b1.debitOperation.reimbursements.length).toBe(2)
+            })
+        })
+      })
+
+      describe('without corresponding credit operation', () => {
```
doubleface: Without corresponding debit operation

ptbrowne:
> One problem remains: the credit operation will be associated to many bills which will have exactly the same pdf file. Should the bank application only display one associated file ?

Yes I agree that it will happen and we will have to deduplicate those in Banks. I think for now we only show the file for the first bill so we'll have to take this into account when adding the functionality.

PR #101, opened 2018-02-26
wip Lazy client

ptbrowne:
It is sometimes a pain to use cozy-konnector-libs because it instantiate the cozy-client directly and the instantiation needs environment variables to be correctly set.

Here, the instantiation is made lazy : we export a proxy to the real client, that is only instantiated when we access `data`, `files`, `jobs` and so on.


ptbrowne:
Yes this is why I put wip :)

ptbrowne:
Too old to be relevant.

PR #102, opened 2018-02-27
Reimbursement filter

ptbrowne:
Added the reimbursement filter so that  you cannot add a bill to an operation
that has already been fully reimbursed.

### DIFF
--- a/libs/linker/billsToOperation/operationsFilters.js
+++ b/libs/linker/billsToOperation/operationsFilters.js
@@ -1,6 +1,7 @@
 const every = require('lodash/every')
 const includes = require('lodash/includes')
 const some = require('lodash/some')
+const sumBy = require('lodash/sumBy')
 const isWithinRange = require('date-fns/is_within_range')
 
 const { getIdentifiers, getDateRangeFromBill, getAmountRangeFromBill } = require('./helpers')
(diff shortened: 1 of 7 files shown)


libs/linker/billsToOperation/operationsFilters.js
```diff
@@ -37,38 +38,71 @@ const isHealthBill = bill => {
...
+  }
+  return reimbursementFilter
 }
 
 // combine filters
 
 const operationsFilters = (bill, operations, options) => {
   const filterByConditions = filters => op => {
-    return every(filters.map(f => f(op)))
+    for (let f of filters) {
```
doubleface: Isn't it the same result ?

package.json
```diff
@@ -50,7 +50,7 @@
   },
   "scripts": {
     "lint": "standard --fix commands/* helpers/* libs/* index.js",
-    "test": "jest",
+    "test": "env LOG_LEVEL=info jest",
```
doubleface: Debug code I suppose

ptbrowne: No, I made the change otherwise the test have a lot of logs

libs/linker/billsToOperation/operationsFilters.js
ptbrowne: With this solution, as soon as there is a false, it returns false. I made this because during debugging, it was easier to see the reasons some bill was matched or not.

PR #107, opened 2018-03-02
typo doc - cozy-client and requestFactory

Benibur:
Troisième tentative...


ptbrowne:
Thanks @Benibur ! The doc is generated automatically from the JS files so your changes would have been overwritten next time. I did the change in https://github.com/cozy/cozy-konnector-libs/pull/108.

ptbrowne:
Actually I will do it in this PR :/

PR #110, opened 2018-03-04
saveFiles handle filestream

### LINKED ISSUE #90: saveFiles should handle file streams as input
This is needed (https://forum.cozy.io/t/developpement-connecteur-avec-factures-html/5326) and it should take the form a fileStream attribute instead of fileUrl I suppose. And it should not be too hard since cozy-client-js already handles file streams.

laedit:
Fix #90 

If the entry has a `filestream` attribute the file is created directly without download.

I just have some remarks:
- I have added the cross-env lib to handle the environment variable in the `test` script, but I could reverse it if necessary
- I throw a new error at line 170, should I create a new error in the `errors.js` file? Or handle it differently?
- At line 113, the error `BAD_DOWNLOADED_FILE` is thrown but since the file isn't always downloaded, should that be modified too?

### DIFF
--- a/libs/saveFiles.js
+++ b/libs/saveFiles.js
@@ -52,25 +52,32 @@ const downloadEntry = function (entry, options) {
     jar: true
   }, entry.requestOptions)
 
+  let filePromise = rq(reqOptions)
+
+  // we have to do this since the result of filePromise is not a stream and cannot be taken by
+  // cozy.files.create
+  if (options.postProcessFile) {
+    return filePromise
+      .then(data => options.postProcessFile(data))
+  }
+  return filePromise
+}
+
+const createFile = function (entry, options) {
   return cozy.files
     .statByPath(options.folderPath)
     .then(folder => {
-      let filePromise = rq(reqOptions)
       const createFileOptions = {
         name: getFileName(entry),
         dirID: folder._id,
         contentType: options.contentType
       }
 
-      // we have to do this since the result of filePromise is not a stream and cannot be taken by
-      // cozy.files.create
-      if (options.postProcessFile) {
-        return filePromise
-        .then(data => options.postProcessFile(data))
-        .then(data => cozy.files.create(data, createFileOptions))
+      if (entry.filestream) {
+        return cozy.files.create(entry.filestream, createFileOptions)
       }
 
-      return cozy.files.create(filePromise, createFileOptions)
+      return cozy.files.create(downloadEntry(entry, options), createFileOptions)
     })
     .then(fileDocument => {
       // This allows us to have the warning message at the first run
@@ -86,7 +93,7 @@ const attachFileToEntry = function (entry, fileDocument) {
 }
 
 const saveEntry = function (entry, options) {
-  if (!entry.fileurl && !entry.requestOptions) return entry
+  if (!entry.fileurl && !entry.requestOptions && !entry.filestream) return entry
 
   if (options.timeout && Date.now() > options.timeout) {
     const remainingTime = Math.floor((options.timeout - Date.now()) / 1000)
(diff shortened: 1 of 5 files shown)


laedit:
My bad, I confused myself when writing the comment, I meant the log at line 133 with the error `FILE_DOWNLOAD_FAILED`.

libs/saveFiles.js
```diff
@@ -52,25 +52,32 @@ const downloadEntry = function (entry, options) {
...
-      if (options.postProcessFile) {
-        return filePromise
-        .then(data => options.postProcessFile(data))
-        .then(data => cozy.files.create(data, createFileOptions))
+      if (entry.filestream) {
+        return cozy.files.create(entry.filestream, createFileOptions)
       }
 
-      return cozy.files.create(filePromise, createFileOptions)
+      return cozy.files.create(downloadEntry(entry, options), createFileOptions)
```
ptbrowne: I think it would be clearer if there was only one `return`.

What do you think of : 

```
const toCreate = entry.filestream || downloadEntry(entry, options)
return cozy.files.create(toCreate, createFileOptions)
```

libs/saveFiles.js
```diff
@@ -86,7 +93,7 @@ const attachFileToEntry = function (entry, fileDocument) {
 }
 
 const saveEntry = function (entry, options) {
-  if (!entry.fileurl && !entry.requestOptions) return entry
+  if (!entry.fileurl && !entry.requestOptions && !entry.filestream) return entry
```
ptbrowne: It starts to be unclear what it is the purpose of this check. I can be
clarified by introducing a function: 

```
const canBeSaved = entry => entry.fileurl || entry.requestOptions || entry.filestream

if (!canBeSaved(entry)) { return entry }
```

libs/saveFiles.spec.js
```diff
@@ -122,4 +130,37 @@ describe('saveFiles', function() {
...
+  })
+
+  const billWithoutStreamUrlAndRequestOptions = [{
+    amount: 62.93,
+    date: '2018-03-03T23:00:00.000Z',
+    vendor: 'Free Mobile',
+    type: 'phone'
+  }]
+  describe('when filestream is used without filename', () => {
+    it('should throw an error', async () => {
```
ptbrowne: I think the text of the assertion is not good ? `it('should do nothing')` maybe ?

libs/saveFiles.js
```diff
@@ -124,7 +131,7 @@ const saveEntry = function (entry, options) {
     })
     .catch(err => {
       log('warn', errors.FILE_DOWNLOAD_FAILED)
-      log('warn', err.message, `Error caught while trying to save the file ${entry.fileurl}`)
+      log('warn', err.message, `Error caught while trying to save the file ${entry.fileurl ? entry.fileurl : entry.filename}`)
```
ptbrowne: SAVE_FILE_FAILED ?

libs/saveFiles.spec.js
laedit: Yes, it is a failed copy/paste, the describe need to be fixed too.

libs/saveFiles.js
laedit: Instead of the `FILE_DOWNLOAD_FAILED`?

libs/saveFiles.js
laedit: Personally, I prefer to return as soon as possible, but it is not my call here :)

libs/saveFiles.js
doubleface: Let's go for SAVE_FILE_FAILED. This error code is not used anywhere else at the moment. Then we can change it freely.

laedit: Done

ptbrowne:
Thanks a lot Jeremie !

PR #113, opened 2018-03-08
🐝 fix: promise check

ptbrowne:
Instead of deleting the check we get the result of fetch inside the
check function instead of returning it in the previous then. It
allow us to check that the result of fetch is a Promise.


ptbrowne:
Related #112

Issue #114, opened 2018-03-08
Allow multiple level error code


doubleface:
Here is the need : have categories of errors. Collect has predefined locales for categories and may have locale for more detailed messages inside categories or not.

Here is the API

```javascript
const {log} = require('cozy-konnector-libs')
log('error', ['LOGIN_FAILED', 'CAPTCHA'])
```

Will result with `LOGIN_FAILED-CAPTCHA` error code ("-" will be forbidden)

If collect knows the `LOGIN_FAILED` code but not `LOGIN-FAILED-CAPTCHA`, it will display the locale associated with `LOGIN_FAILED`.

ptbrowne:
I would use the dot instead of the dash. "LOGIN_FAILED.CAPTCHA" indicates better that one is the child of the other.

PR #117, opened 2018-03-09
feat : convert critical array message to format expected by collect

doubleface:
Implements #114

### DIFF
--- a/libs/log-formats.js
+++ b/libs/log-formats.js
@@ -21,7 +21,27 @@ function prodFormat (type, message, label, namespace) {
   if (message.stack) message = message.stack
 
   // cut the string to avoid a fail in the stack
-  return JSON.stringify({ time: new Date(), type, message, label, namespace }).substr(0, LOG_LENGTH_LIMIT)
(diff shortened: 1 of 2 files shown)


libs/log-formats.js
```diff
@@ -21,7 +21,27 @@ function prodFormat (type, message, label, namespace) {
...
+  if (type !== 'critical') return message
+
+  let result = message
+  if (Array.isArray(result)) {
+    // clean non string items and remove '-' in items
+    result = result
+      .filter(item => typeof item === 'string')
+      .map(item => item.replace(/-/g, ''))
+      .join('-')
+  }
```
ptbrowne: I am not sure this is the place for this. As stated in #114 , I am not a fan of the API with an Array, I find it clear enough with "LOGIN_FAILED.CAPTCHA". If we keep it simple, we do not have to add logic to the logger to handle this case.

doubleface: I think you are right @ptbrowne . I will let this PR for a moment and close it if no other advice comes.

PR #118, opened 2018-03-09
Update cozy-client

ptbrowne:
cozy-client had not been updated in almost a year :/


ptbrowne:
Will be done via renovate

PR #125, opened 2018-03-09
✨ feat: add scrape method for declarative scraping

ptbrowne:
The scrape method allows declarative scraping.

Declarative scraping is good at it could allow us to build interfaces for people to more easily build a konnector.

```js
import { scrape } from 'cozy-konnector-libs'

const $ = cheerio.load(html)
scrape($, {
  title: '.title',
  content: '.content',
  link: {
    sel: 'a',
    attr: 'href',
    index: 0
  }
})
// -> {
//   title: 'Adam Ondra',
//   content: 'Adam Ondra is the first climber to ever climb 9C.',
//   link: 'https://fr.wikipedia.org/wiki/Adam_Ondra'
// }
"
```

It is heavily inspired (almost copied the API) by [artoo].

It also allows to scrape several items in one go via its third parameter. 

```js
scrape($, { title: '.title'}, '.articles')
// -> [{ title: 'a' }, { title: 'b'}, { title: 'c'}]
```

Check the tests ;)

[artoo]: https://medialab.github.io/artoo/scrape/#scrape

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,8 @@ module.exports = {
   requestFactory,
   retry: require('bluebird-retry'),
   wrapIfSentrySetUp: require('./helpers/sentry').wrapIfSentrySetUp,
-  Document: require('./libs/document')
+  Document: require('./libs/document'),
+  scrape: require('./libs/scrape')
 }
 
 function deprecate (wrapped, message) {
--- a/libs/scrape.js
+++ b/libs/scrape.js
@@ -1,0 +1,67 @@
+/**
+ * Declarative scraping.
+ *
+ * Describe your items attributes and where to find/parse them
+ * instead of imperatively building them.
+ *
+ * Heavily inspired by [artoo] scraping method.
+ *
+ * [artoo]: https://medialab.github.io/artoo/
+ */
+
+const mkSpec = function(spec) {
+  if (typeof spec === "string") {
+    return { sel: spec };
+  } else {
+    return spec;
+  }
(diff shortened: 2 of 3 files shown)


doubleface:
Awesome! I did not know artoo!
I made a branch in [cozy-konnector-template](https://github.com/doubleface/cozy-konnector-template/blob/feat/scrape/index.js) using scrape and I got a problem : the connector does not finish and takes all the processor using yarn standalone. This is a problem I already had.

If I replace : 

```javascript
    return ($.find || $)(childSelector).map((i, e) => scrape($(e), specs));
```

with : 

```javascript
    return Array.from(($.find || $)(childSelector)).map(e => scrape($(e), specs));
```

The problem disappears and I have no clear idea why.

Another thing, in the template, when I do : 

```javascript
      filename: {
        sel: 'h3 a',
        fn: $el => `${$el.attr('title')}.jpg`
      }
```

I would prefer to do : 

```javascript
      filename: {
        sel: 'h3 a',
        attr: 'title',
        fn: title => `${title}.jpg`
      }
```
What do you think?

ptbrowne:
- `Array.from` fixed, thanks !

For your second question, you can already do : 

```
      filename: {
        sel: 'h3 a',
        attr: 'title',
        parse: title => `${title}.jpg`
      }
```

In other words, `fn` is to do custom stuff with the node, `parse` is to do custom stuff with the value extracted.

PR #128, opened 2018-03-12
Doc for log levels: warning → warn

carrieje:
https://github.com/cozy/cozy-konnector-libs/blob/master/libs/log-formats.js#L11


ptbrowne:
Thanks !

PR #137, opened 2018-03-14
Provide test for request.js

carrieje:
Test encoding option for `requestFactory`.

I don't know if you will find it useful.
I have written them to get used to writing tests in your libs.

I am still unable to make `replay` work in your environment.
These tests emit requests, which make them network dependent.
If no one can patch them with `replay`, the network dependency can be annoying.


kosssi:
thanks a lot 👍

Issue #136, opened 2018-03-13
Standalone is no longer working in last build


carrieje:
`myrepo/package.json`
```js
   "dependencies": {
-    "cozy-konnector-libs": "^3.2.5",
+    "cozy-konnector-libs": "^3.7.2",
+    "cozy-jobs-cli": "^1.0.3",
+    "cozy-logger": "^1.0.3",
     "csv-parse": "^2.0.4",
     "moment-timezone": "^0.5.14"
   },
   "scripts": {
     "start": "node src/index.js",
-    "dev": "cozy-konnector-dev src/index.js",
-    "standalone": "cozy-konnector-standalone src/index.js",
-    "standalone:record": "REPLAY=record cozy-konnector-standalone src/index.js",
-    "standalone:replay": "REPLAY=replay cozy-konnector-standalone src/index.js",
+    "dev": "cozy-run-dev src/index.js",
+    "standalone": "cozy-run-standalone src/index.js",
+    "standalone:record": "REPLAY=record cozy-run-standalone src/index.js",
+    "standalone:replay": "REPLAY=replay cozy-run-standalone src/index.js",
```

Running `yarn standalone` lead to this error:

```sh
Error: Cannot find module 'cozy-konnector-libs/helpers/init-konnector-config'
[…]
    at Object.<anonymous> (<MYREPO>/node_modules/cozy-jobs-cli/src/standalone.js:17:16)
```

This is fixed by patchin `cozy-jobs-cli/src/standalone.js`

```js
17- const config = require('cozy-konnector-libs/helpers/init-konnector-config')()
17+ const config = require('./init-konnector-config')()                              
```

And  it gets stuck right after:

```js
Error: Cannot find module './init-konnector-config'
[…]
    at Object.find (<myrepo>/node_modules/cozy-konnector-libs/helpers/cozy-client-js-stub.js:57:24)
```

This line is responsible for this:

```js
        const config = require('./init-konnector-config')()
```

I don't know `lerna`. I don't know if there is a way to share this same file between the two packages.

ptbrowne:
Thanks for signaling this to us, I saw this problem this morning. I'm on it. Sorry you encountered this problem, I should have done a prerelease :/

PR #138, opened 2018-03-14
Provide signon helper to the user

carrieje:
The [carrieje/cozy-konnector-connection](https://github.com/carrieje/cozy-konnector-connection) seems to fit [@laedit](https://forum.cozy.io/t/nouveau-module-konnector-connection/5517/7) [expectations](https://github.com/laedit/cozy-konnector-theoldreader/blob/master/index.ts#L33-L36).
I have used it in three connectors successfully.

@doubleface spoke about [integrating it to the libs](https://forum.cozy.io/t/nouveau-module-konnector-connection/5517/3).
Here is my PR for this to be done.

I have not written tests, as ( #137 ) I am not able to do it using `replay`.
I have not yet tested my contribution manually either, as ( #136 ) I would wait for standalone to work again.

I can at least wait for #136 to be resolved, and test it manually before you take a look.

I have made the choice to name it `signon` as I expect people to integrate it to their `login` function.
And I did not want to reserve the `login` name.
This choice is debatable.

### DIFF
--- a/packages/cozy-konnector-libs/index.js
+++ b/packages/cozy-konnector-libs/index.js
@@ -21,6 +21,7 @@ module.exports = {
   retry: require('bluebird-retry'),
   wrapIfSentrySetUp: require('./helpers/sentry').wrapIfSentrySetUp,
   Document: require('./libs/document'),
+  signon: require('./libs/signon'),
   scrape: require('./libs/scrape')
 }
 
--- a/packages/cozy-konnector-libs/index.js
+++ b/packages/cozy-konnector-libs/index.js
@@ -21,7 +21,7 @@ module.exports = {
   retry: require('bluebird-retry'),
   wrapIfSentrySetUp: require('./helpers/sentry').wrapIfSentrySetUp,
   Document: require('./libs/document'),
-  signon: require('./libs/signon'),
+  signin: require('./libs/signin'),
   scrape: require('./libs/scrape')
 }
 
--- a/packages/cozy-konnector-libs/libs/signin.js
+++ b/packages/cozy-konnector-libs/libs/signin.js
@@ -53,17 +53,17 @@ module.exports = function signin (
   validate = defaultValidate,
   opts = {})
 {
-  const defaultOpts = { jar: true, cheerio: true, json: false }
-
   const rq = requestFactory({
-    ...opts,
-    ...defaultOpts
+    jar: true,
+    ...opts
   })
 
   const parseBody = getStrategy(parse)
 
-  return rq(pageUrl)
-    .catch(handleRequestErrors)
+  return rq({
+    uri: pageUrl,
+    cheerio: true
+  }).catch(handleRequestErrors)
     .then($ => {
       const [action, inputs] = parseForm($, formSelector)
       for (let name in formData) {
--- a/packages/cozy-konnector-libs/libs/signin.js
+++ b/packages/cozy-konnector-libs/libs/signin.js
@@ -53,7 +53,7 @@ module.exports = function signin (
     ...defaultOpts
   })
 
-  const parseBody = defineStrategy(parseStrategy)
+  const parseBody = getStrategy(parseStrategy)
 
   return rq(`${baseUrl}/${page}`)
   .then($ => {
@@ -77,7 +77,7 @@ function defaultValidate (statusCode, body) {
   return statusCode === 200
 }
 
-function defineStrategy (parseStrategy) {
+function getStrategy (parseStrategy) {
   switch (parseStrategy) {
     case 'cheerio':
       return require('cheerio').load
--- a/packages/cozy-konnector-libs/libs/signin.js
+++ b/packages/cozy-konnector-libs/libs/signin.js
@@ -4,7 +4,7 @@
  * failure, it throws a `LOGIN_FAILED` error.
  *
  * It does not submit values provided through `select` tags, except if populated
- * by user with `population`.
+ * by user with `formData`.
  * Limit between `baseUrl` and `page` is important. The form `action` will be
  * concatenated the same way `page` is, to constitute the submission url. For
  * now, a form with an absolute `action` is not comptabile with this function.
@@ -14,7 +14,7 @@
  * - `formSelector` is used by cheerio to uniquely identify the form in which to
  *   log in
  *
- * - `population` is an object of { name: value, … }. It is used to populate the
+ * - `formData` is an object of { name: value, … }. It is used to populate the
  *   form, in the proper inputs with the same name as the properties of this
  *   object, before submitting it.
  *
@@ -39,7 +39,7 @@ const url = require('url')
 module.exports = function signin (
   pageUrl,
   formSelector,
-  population,
+  formData,
   parseStrategy = 'cheerio',
   validate = defaultValidate,
   opts = {}) {
@@ -56,8 +56,8 @@ module.exports = function signin (
   return rq(pageUrl)
   .then($ => {
     const [action, inputs] = parseForm($, formSelector)
-    for (let name in population) {
-      inputs[name] = population[name]
+    for (let name in formData) {
+      inputs[name] = formData[name]
     }
 
     return submitForm(rq, url.resolve(pageUrl, action), inputs, parseBody)
--- a/packages/cozy-konnector-libs/libs/signin.js
+++ b/packages/cozy-konnector-libs/libs/signin.js
@@ -34,7 +34,7 @@
  *   `parsedBody`. If it is false, `LOGIN_FAILED` is thrown, otherwise the
  *   konnector continues.
  *
- * - `opts` allows to pass eventual options to the `signin`'s `requestFactory`.
+ * - `requestOpts` allows to pass eventual options to the `signin`'s `requestFactory`.
  *   It could be useful for pages using `latin1` `encoding` for instance.
  *
  * @module signin
@@ -52,12 +52,12 @@ module.exports = function signin (
   {
     parse = 'cheerio',
     validate = defaultValidate,
-    ...opts
+    ...requestOpts
   } = {})
 {
   const rq = requestFactory({
     jar: true,
-    ...opts
+    ...requestOpt
   })
 
   const parseBody = getStrategy(parse)
--- a/packages/cozy-konnector-libs/libs/signin.js
+++ b/packages/cozy-konnector-libs/libs/signin.js
@@ -54,6 +54,14 @@ module.exports = function signin (
     ...requestOpts
   } = {})
 {
+  // Check for mandatory arguments
+  if (url === undefined) {
+    throw 'signin: `url` must be defined'
+  }
+  if (formSelector === undefined) {
+    throw 'signin: `formSelector` must be defined'
+  }
+
   const rq = requestFactory({
     jar: true,
     ...requestOpts
--- a/packages/cozy-konnector-libs/libs/signin.js
+++ b/packages/cozy-konnector-libs/libs/signin.js
@@ -32,6 +32,7 @@
  * @module signin
  */
 const errors = require('../helpers/errors')
+const rerrors = require('request-promise/errors')
 const log = require('cozy-logger').namespace('cozy-konnector-libs')
 const requestFactory = require('./request')
 const url = require('url')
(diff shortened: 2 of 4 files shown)


carrieje:
I chose not to commit the `yarn doc` output as it seems to add documentation for other things than my sole module.

carrieje:
Applied #140 locally.
I have tried my PR by manually editing the `node_modules` files.
It is fixed and have proven to be working.

packages/cozy-konnector-libs/libs/signon.js
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @module signon
+ */
+const errors = require('../helpers/errors')
+const log = require('cozy-logger').namespace('cozy-konnector-libs')
+const requestFactory = require('./request')
+
+module.exports = function signon (
+  baseUrl,
+  page,
```
ptbrowne: I think it would be simpler if there was only 1 url in the signature : baseUrl + page. I think the concatenation should not be the responsibility of this function.

packages/cozy-konnector-libs/libs/signon.js
```diff
@@ -0,0 +1,114 @@
...
+  formSelector,
+  population,
```
ptbrowne: I would rename population into formAttrs for form attributes, or formData.

packages/cozy-konnector-libs/libs/signon.js
```diff
@@ -0,0 +1,114 @@
...
+  const parseBody = defineStrategy(parseStrategy)
+
+  return rq(`${baseUrl}/${page}`)
+  .then($ => {
+    const [action, inputs] = parseForm($, formSelector)
+    for (let name in population) {
+      inputs[name] = population[name]
+    }
+
+    return submitForm(rq, `${baseUrl}/${action}`, inputs, parseBody)
```
ptbrowne: OK, I understand better why baseUrl was separated from the page. Still, I think you can find the baseUrl from the page using a url parser. This will be more robust. I imagine a case where we could fail otherwise.

sign in page is http://amazon.com/login/login.html
action is : http://amazon.com/login/signin written in the form as action='signin`

the developer uses 

```
signin('http://amazon.com', '/login/login.html')
```

thinking that baseUrl means the root url of the site.

Here the action url will be 'http://amazon.comsignin' which is not what we want. This is I would prefer if the user only had to pass the absolute page URL and then we convert relative URLs to absolute URLs with the knowledge of the URL we landed on. I think it will also be more robust in the face of redirections.

packages/cozy-konnector-libs/libs/signon.js
```diff
@@ -0,0 +1,114 @@
...
+      return Promise.resolve(parsedBody)
+    }
+  })
+}
+
+function defaultValidate (statusCode, body) {
+  return statusCode === 200
+}
+
+function defineStrategy (parseStrategy) {
```
ptbrowne: At first I thought that `define` meant that there would be a side-effect.

I would write `getStrategy(name)`.

packages/cozy-konnector-libs/libs/signon.js
```diff
@@ -0,0 +1,114 @@
...
+  switch (parseStrategy) {
+    case 'cheerio':
+      return require('cheerio').load
+    case 'json':
+      return JSON.parse
+    default:
+      let err = `connection: parsing strategy ${parseStrategy} unknown. `
+      let fallback = 'Falling back to `raw`. Use one of `raw`, `cheerio` or `json`'
+      log('warn', err + fallback)
```
ptbrowne: Since it seems to be an error from the developer, I would prefer to fail fast on this, without fallback.

packages/cozy-konnector-libs/libs/signon.js
```diff
@@ -0,0 +1,114 @@
...
+  })
+
+  const parseBody = defineStrategy(parseStrategy)
+
+  return rq(`${baseUrl}/${page}`)
+  .then($ => {
+    const [action, inputs] = parseForm($, formSelector)
+    for (let name in population) {
+      inputs[name] = population[name]
+    }
```
ptbrowne: I wonder if this part could not be merged with parseForm ?

packages/cozy-konnector-libs/index.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
   retry: require('bluebird-retry'),
   wrapIfSentrySetUp: require('./helpers/sentry').wrapIfSentrySetUp,
   Document: require('./libs/document'),
+  signon: require('./libs/signon'),
```
ptbrowne: signon -> signin ?

packages/cozy-konnector-libs/libs/signon.js
carrieje: If I simplify to a single loop (on `input` names), the user will not be able to define extra parameters.
For now, this is the only way it has to add `select` values to the form, as `parseForm` does not care for these elements.

If we want to keep the ability to manage `select` inputs, and without a clean way to handle them, moving this code logic to `parseForm` would just relocate the second loop.

Although, I am not against relocating it :-)

carrieje:
I would like to train myself writing tests in js.
I will notify you when I have written it. 🙂 
You will be free to write it from scratch if it does not suit you 😄

carrieje:
@ptbrowne Here it is ;-)
I would be glad to get your reviews
Don't hesitate to tell if you expected some things to be different.

packages/cozy-konnector-libs/libs/signin.js
```diff
@@ -1,7 +1,15 @@
 /**
  * The goal of this function is to provide an handy method to log the user in,
- * on html form pages. On success, it resolves a promise with a parsed body. On
- * failure, it throws a `LOGIN_FAILED` error.
+ * on html form pages. On success, it resolves a promise with a parsed body.
+ *
+ * Errors:
+ *
+ * - LOGIN_FAILED if the validate predicate is false
+ * - INVALID_FORM if the element matched by `formSelector` is not a form or has
```
carrieje: I was wondering if I should throw a `LOGIN_FAILED` with an explicit log error message about the `form` not matching basic needs. WDYT ?
It would be more robust to layout changes by the provider.

carrieje:
It's ok for me.
I have finished handling `StatusCodeError`.

Have a good weekend ;-)

carrieje:
It's ok for me.
I have finished handling `StatusCodeError`.

Have a good weekend ;-)

packages/cozy-konnector-libs/libs/signin.js
```diff
@@ -0,0 +1,145 @@
...
+const requestFactory = require('./request')
+const url = require('url')
+
+module.exports = function signin (
+  pageUrl,
+  formSelector,
+  formData,
+  parseStrategy = 'cheerio',
+  validate = defaultValidate,
+  opts = {}) {
```
ptbrowne: When looking at the tests, I was wondering what were each arguments, I think it would be clearer if signin would take an option object. What do you think ?

carrieje: Do you mean something like this?

```js
function signin(pageUrl, formSelector, {formData = {}, parseStrategy = 'cheerio', validate = defaultValidate, ...opts})
```

packages/cozy-konnector-libs/libs/signin.js
```diff
@@ -0,0 +1,145 @@
...
+  const defaultOpts = { jar: true, cheerio: true, json: false }
+
+  const rq = requestFactory({
+    ...opts,
+    ...defaultOpts
```
ptbrowne: `...opts` and `...defaultOts` should be switched as here it is not possible to overwrite a default option.

packages/cozy-konnector-libs/libs/signin.js
ptbrowne: > It would be more robust to layout changes by the provider.

I don't understand, why "INVALID_FORM" would be more robust than "LOGIN_FAILED" ?

I think it is better like with "INVALID_FORM". LOGIN_FAILED would be interpreted by the connection form (the collect application) and a message would be displayed to the user saying that its credentials are invalid, and that would be wrong.

packages/cozy-konnector-libs/libs/signin.js
ptbrowne: More like `signin({ pageUrl, formSelector, formData, parseStrategy, validate})`.

Also by writing this, I realized that parseStrategy and validate are both strategies but one has "strategy" and the other not, it would be better if they had the same naming ~~strategy~~ scheme. IMHO without "strategy" is better.

ptbrowne:
Again thanks a lot for your patience and contribution Jeremy ! I am sorry for suggestions that may seem minor but since it will be a public API, I prefer to suggest now before it is used and having to change it later. I hope you're OK with this :)

packages/cozy-konnector-libs/libs/signin.js
carrieje: It is made on purpose, as defaultOpts contains `cheerio: true` and that this behavior is needed for core implementation of `signin`.
Everything that is un-overwritteable by `defaultOpts` can be customized for the output.

If you want output as `json` you switch your parsing strategy to `json`.

I agree on one thing though, if you want to provide a specific `jar` you won't be able to do it.

I suggest the following:

```js
const rq = requestFactory({
  jar: true,
  ...opts,
  cheerio: true,
  json: false
})
```

WDYT?

packages/cozy-konnector-libs/libs/signin.js
carrieje: AYW.
The thing is that by doing so, you lose the information on which argument is mandatory.
You cannot signin without giving an URL and a form selector. Hence, the suggestion I made.

If `pageUrl` and `formSelector` are in the object, we would have to write code to ensure they are present.

I'm ok with the naming ;-)

carrieje:
I'm perfectly OK with it. Don't worry :-)
Even if I give argumentations on your remarks (which I hope this is what you expect) you would have the last word on what to do. :-)
And I'm ok with that too.

packages/cozy-konnector-libs/libs/signin.js
carrieje: I'm ok with your interpretation then.
I had the idea that LOGIN_FAILED would have a broader interpretation than wrong credentials.

Would we need to add the INVALID_FORM to the main errors documentation of the repo if we keep it like this?

packages/cozy-konnector-libs/libs/signin.js
carrieje: I commit this then ;-)

packages/cozy-konnector-libs/libs/signin.js
```diff
@@ -0,0 +1,146 @@
...
+const requestFactory = require('./request')
+const url = require('url')
+
+module.exports = function signin (
+  pageUrl,
+  formSelector,
+  formData,
+  parse = 'cheerio',
+  validate = defaultValidate,
+  opts = {})
```
doubleface: `requestOpts` would be more explicit

And in the best case I would get : 

```javascript
signin('https://http://quotes.toscrape.com/login', 'form', {username: 'test', password: 'passw0rd'})
```

Still, I prefer named parameters with an option object.

Handling default values and mandatory parameters would be less straightforward but I think the connector developers would find it easier to use.

carrieje: > requestOpts would be more explicit

92b36d5

> Still, I prefer named parameters with an option object.

So are you ok with 026d748 ?

doubleface: Why not , from connector developer point of view : 

```javascript
signin(
  url: 'https://http://quotes.toscrape.com/login',
  formSelector: 'form',
  formData: {
    username: 'test',
    password: 'passw0rd'
})
```

?

carrieje: If I am not wrong, he would have to wrap everything in an object like so:

```js
signin({
  url: 'https://http://quotes.toscrape.com/login',
  formSelector: 'form',
  formData: {
    username: 'test',
    password: 'passw0rd'
  }
})
```

If you confirm this is what you want, I will commit tomorrow ;-)

carrieje: 0b8009a

doubleface:
Let's release it and test it with our next connector

PR #139, opened 2018-03-14
Fix doc script

carrieje:
Was broken by repository reorganisation


ptbrowne:
Thanks, we still have some cleaning up to do after the reorg...

PR #142, opened 2018-03-14
fix: add yarn.lock + cozy-logger dependency in cozy-konnector-libs

doubleface:


doubleface:
I did not see the cleanup PR

PR #149, opened 2018-03-19
Link visualizer

ptbrowne:
/!\ Depends on #148

Introduces a server to visualize link between bank transactions and bills.
It is useful for debugging link between operations. It should be launched
with `cozy-run-dev` so in theory it could be run against a remote cozy.

To be able to do that, I had to :

* Do the linking in memory, we fetch all the documents from the DB and do the
filtering in memory instead of asking couch to do it for us. It is much more quick
as we do not have to do ${number of bills} queries.

* Separate the finding-the-links operations from the saving-the-links operation. It
allows to subclass the Linker in the visualizing server to replace it saving-the-links
(represented by the commitChanges function) by a dummy function.

It is still in WIP as I did not put too much tests and want to test more thoroughly
the changes I made to the linking logic.


packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -64,21 +66,63 @@ class Linker {
...
+    this.toUpdate.push(doc)
+    return Promise.resolve()
+  }
+
+  commitChanges () {
+    return cozyClient.fetchJSON(
+      'POST',
+      `data/${DOCTYPE_OPERATIONS}/_bulk_docs`,
+      {
+        data: JSON.stringify
```
drazik: Is it normal that you're passing a function to `data` ?

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -64,21 +66,63 @@ class Linker {
...
+      options.identifiers = options.identifiers.map(id => id.toLowerCase())
+    } else {
+      throw new Error(
+        'linkBankOperations cannot be called without "identifiers" option'
+      )
+    }
+
+    options.amountDelta = options.amountDelta || DEFAULT_AMOUNT_DELTA
+    options.minAmountDelta = options.minAmountDelta || options.amountDelta
+    options.maxAmountDelta = options.maxAmountDelta || options.amountDelta
```
drazik: Can `options.amountDelta`, `options.minAmountDelta` or `options.maxAmountDelta` be `0` ? If yes, you may end with `options.minAmountDelta === options.amountDelta` where you wanted it to be `0` initially.

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -87,9 +131,10 @@ class Linker {
       const res = result[bill._id] = { bill:bill }
 
       const linkBillToDebitOperation = () => {
-        return findDebitOperation(this.cozyClient, bill, options)
+        return findDebitOperation(this.cozyClient, bill, options, allOperations)
           .then(operation => {
             if (operation) {
+              console.log('found debit operation', operation)
```
drazik: Is it a temporary `console.log` here ? If not, maybe use `log` ?

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -99,10 +144,11 @@ class Linker {
       }
 
       const linkBillToCreditOperation = debitOperation => {
-        return findCreditOperation(this.cozyClient, bill, options)
+        return findCreditOperation(this.cozyClient, bill, options, allOperations)
           .then(creditOperation => {
             const promises = []
             if (creditOperation) {
+              console.log('found credit operation')
```
drazik: Same here ?

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -133,31 +179,16 @@ const jsonTee = filename => res => {
...
-
-  options.pastWindow = options.pastWindow || DEFAULT_PAST_WINDOW
-  options.futureWindow = options.futureWindow || DEFAULT_FUTURE_WINDOW
-
+  // Use the custom bank identifier from user if any
   const cozyClient = require('./cozyclient')
   const linker = new Linker(cozyClient)
   const prom = linker.linkBillsToOperations(bills, options)
+    .catch(err => {
+      console.log(err)
```
drazik: Same question about `console.log` here

packages/cozy-konnector-libs/libs/linker/billsToOperation/findNeighboringOperations.js
```diff
@@ -0,0 +1,89 @@
...
+      return and(conditions)(doc)
+    })
+  return docs.filter(and(filters))
+}
+
+const findNeighboringOperations = (cozyClient, bill, options, allOperations) => {
+  let operations = []
+
+  const queryOptions = getQueryOptions(bill, options, [])
+  // allOperations = sortBy(queryOptions.sort)
```
drazik: Maybe a line to remove ?

packages/cozy-konnector-libs/libs/linkBankOperations.js
ptbrowne: Thanks for the catch

packages/cozy-konnector-libs/libs/linkBankOperations.js
ptbrowne: thx

packages/cozy-konnector-libs/libs/linkBankOperations.js
ptbrowne: thx

packages/cozy-konnector-libs/libs/linker/billsToOperation/findNeighboringOperations.js
ptbrowne: yes

packages/cozy-konnector-libs/libs/linkBankOperations.js
ptbrowne: I used defaults from lodash since `options.minAmountDelta = options.minAmountDelta !== undefined ? options.minAmountDelta : DEFAULT_MIN_AMOUNT_DELTA` is too verbose for my taste.

ptbrowne:
> Maybe I missed something, but I don't understand how I can start a server to use this tool. I think a little bit of documentation would be great when your work on it is done 👍

Très bon point, désolé, j'aurais du prendre le temps de faire ça un peu plus tôt.

packages/cozy-konnector-libs/libs/utils.js
```diff
@@ -0,0 +1,12 @@
+const cozyClient = require('./cozyclient')
+
+const fetchAll = async doctype => {
```
doubleface: You could use `cozy.client.data.findAll(doctype)` https://cozy.github.io/cozy-client-js/data-api.html

ptbrowne:
Merged in https://github.com/konnectors/libs/commit/f10b33e3e37c6fb07896b77ecf77aa9213d1ba21

PR #152, opened 2018-03-21
🐝 fix: check all health categories

ptbrowne:
Check all health categories since Linxo seems to have change its
categorization of some insurances from "healthExpenses" to "healthInsurance".

```
   "400600": "health",
   "400610": "healthExpenses",
   "400620": "healthInsurance",
```


doubleface:
I think we should not match health expenses with the wrong category.

We have a case here were linxo gives a wrong category and we should not compensate this mistake with another mistake in the matcher (taking other health categories operation for the matching)

fffflo:
I would naturally agree with doubleface : "others" allow users to put some debit in this section so that they can be excluded from the matching. 
"healthinsurance" is dedicated to the payment of the insurance. 
@ptbrowne have you come across a lot of cases where the health expenses was in "healthinsurance" ?  if it the case we can look more into it (and ask linxo first) but I would continue to exclude "others"

ptbrowne:
The cases I stumbled upon were Malakoff ederic reimbursements. For now we only match health bills (identified by the presence of "Harmonie" or "Malakoff Mederic" etc.. in their label), with health operations (identified by 400610 aka Health expense). The problem is that reimbursements from insurances are not in the 400610 category but in the Health insurance.

What would be better would be to allow reimbursements (credit operations) to be in Health insurance category and expenses to be in Health expenses category.

fffflo:
image
When I take a look at linxo UI, they use 400620 for the subscription plan ("cotisation mutuelle prévoyance") and 400610 is supposedly only for actual health expenses and reimbursments "frais, remb santé"

fffflo:
it is possible to enlarge our matching to cover for linxo mistakes : we should only look at 400620 when looking for the reimbursment (matching with the credit). 
we should not look in the 400620 when looking for the expense (debit matching) otherwise we will include subscription plans in our potential "matchs".

ptbrowne:
Sounds good, I'll update the PR.

PR #161, opened 2018-03-28
feat: Launch renovate only one time by week 🍔

kosssi:


ptbrowne:
I would be in favor of letting renovate run for cozy-konnector-libs as soon as it is published.

rarkins:
I think it should work. I recommend this change though:
```
      "extends": [
        "config:base"
      ],
      "timezone": "Europe/Paris",
      "minor": {
        "automerge": true
      },
      "packageRules": [{
        "excludePackageNames": ["cozy-konnector-libs"],
        "schedule": "before 5am every monday",
      }],
      "prHourlyLimit": 0
```

If you are raising them all once a week, best to remove the hourly rate limiting so that you don't run out of time to get all PRs created.

kosssi:
thanks a lot for your responce

PR #168, opened 2018-04-04
feat(cozy-konnector-libs): match grouped debit

drazik:
The process is the following : 

* We try to match each bill with some operations just as we did before
* After that, we take all the bills that don't have a matching debit operation
* Then we group them on their `originalDate` property
* Then we compute all the possible combinations for each group
* Then for each possible combination, we create a fake bill that sums the `amount` and `originalAmount` of the bills in the combination
* Then we try to match a debit operation for these fake bills
* Finally, if we found a debit operation for the fake bill, we link all the bills that compose the fake bill to this debit operation

I tested this in the following cases : 

image

With this result : 
image

And 
image

With this result : 
image

I'm just wondering if it would not be good to group the bills on `originalDate` AND `type`, to be sure to not combine totally unrelated bills. What do you think ?


packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -13,6 +13,9 @@ const { findDebitOperation, findCreditOperation } = require('./linker/billsToOpe
 const fs = require('fs')
 const { fetchAll } = require('./utils')
 const defaults = require('lodash/defaults')
+const groupBy = require('lodash/groupBy')
+const flatten = require('lodash/flatten')
```
ptbrowne: <3

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -13,6 +13,9 @@ const { findDebitOperation, findCreditOperation } = require('./linker/billsToOpe
...
+const geco = require('geco')
```
ptbrowne: https://www.npmjs.com/package/geco

Bringing the heavy artillery.

> It is extremely memory efficient and very fast, it generates all the k-combinations in a set of n elements, using a single sequence of n bits, only 1 bit per element.

> - total: 2598960 combinations
> - rate:  621760 comb/sec
> - time:  4.18 secs

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -170,9 +173,83 @@ class Linker {
       })
     })
     .then(() => {
-      return result
+      const notLinkedBills = this.getNotLinkedBills(result)
```
ptbrowne: Unlinked ?

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -170,9 +173,83 @@ class Linker {
...
+          .values(billsGroups)
+          .map(billsGroup => this.generateBillsCombinations(billsGroup))
+      )
+
+      const combinedBills = combinations.map(
+        combination => this.combineBills(...combination)
+      )
+
+      return Promise.all(combinedBills.map(bill => {
+        return findDebitOperation(this.cozyClient, bill, options, allOperations)
```
ptbrowne: findDebitOperation is synchronous now ? Maybe we could get rid of all this Promise handling ?

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -170,9 +173,83 @@ class Linker {
...
+      }))
+        .then(debitOperations => debitOperations.filter(Boolean))
+        .then(debitOperations => {
+          debitOperations.forEach((debitOperation, index) => {
+            combinedBills[index].originalBills.forEach(originalBill => {
```
ptbrowne: Why is not `bill` here ? We are indexing bills with an index from operations ??

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -170,9 +173,83 @@ class Linker {
...
+        for (const combination of iter) {
+          newCombinations.push(combination)
+        }
+
+        return [
+          ...combinations,
+          ...newCombinations
+        ]
+    }, [])
+
```
ptbrowne: It seems a bit complicated.

Why not simply do 

```
let combinations = []
for (let n = 0; n < combinations.length; n++) {
	const combinationsN = geco.gen(bills.length, n + 1, bills)
    combinations = combinations.concat(combinationsN)
}
return combinations
```

This way you do not have the need for the flatten above.

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -170,9 +173,83 @@ class Linker {
...
+    return combinations
+  }
+
+  combineBills (...bills) {
+    return {
+      ...bills[0],
+      _id: ['combined', ...bills.map(bill => bill._id)].join(':'),
+      amount: bills.reduce((sum, bill) => sum + bill.amount, 0),
+      originalAmount: bills.reduce((sum, bill) => sum + bill.originalAmount, 0),
```
ptbrowne: minor: lodash has `sumBy(bills, bill => bill.originalAmount)`

packages/cozy-konnector-libs/libs/linkBankOperations.spec.js
```diff
@@ -285,4 +285,110 @@ describe('linker', () => {
       })
     })
   })
+
+  describe('linking with combinations', () => {
+    describe('getNotLinkedBills', () => {
+
+      test('it returns the bills that are not linked', () => {
```
ptbrowne: `it('returns the bills that are not linked')`


`it` can be conveniently into `fit` which will focus jest on only this test.

packages/cozy-konnector-libs/libs/linkBankOperations.spec.js
```diff
@@ -285,4 +285,110 @@ describe('linker', () => {
...
+        { _id: 'b4', originalDate: '2018-03-15T00:00:00Z' },
+        { _id: 'b5', originalDate: '2018-03-15T00:00:00Z' },
+        { _id: 'b6', originalDate: '2018-03-20T00:00:00Z' },
+        { _id: 'b7', originalDate: '2018-03-20T00:00:00Z' },
+        { _id: 'b8', originalDate: '2018-03-20T00:00:00Z' },
+        { _id: 'b9', originalDate: '2018-03-20T00:00:00Z' },
+        { _id: 'b10', originalDate: '2018-03-30T00:00:00Z' }
+      ]
+
+      test('it groups bills by their originalDate property', () => {
```
ptbrowne: I think the time should be removed from the key. It can cause some problems because of timezone etc..

packages/cozy-konnector-libs/libs/linkBankOperations.spec.js
```diff
@@ -285,4 +285,110 @@ describe('linker', () => {
...
+        const combinedBill = linker.combineBills(...bills)
+        expect(combinedBill.originalAmount).toBe(30)
+      })
+
+      test('it generates a bill with the right originalDate', () => {
+        const combinedBill = linker.combineBills(...bills)
+        expect(combinedBill.originalDate).toBe('2018-03-10T00:00:00Z')
+      })
+    })
+  })
```
ptbrowne: There are no tests that tries to link with groups (more like integration tests). I think it should be good to add them. They will be more robust that the unit tests.

drazik:
Is it better now for you @ptbrowne ?

packages/cozy-konnector-libs/scripts/visualizer/index.js
```diff
@@ -53,7 +52,7 @@ const parseOptions = parser({
   allowUncategorized: htmlFormBoolean,
   minDateDelta: float,
   maxDateDelta: float,
-  amountDelta: float,
+  amountDateDelta: float,
```
kosssi: C'est bizarre ce changement non ?

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -174,40 +174,49 @@ class Linker {
...
             log('debug', combinedBill, 'Matching bills combination')
             log('debug', debitOperation, 'Matching debit debitOperation')
 
-            combinedBill.originalBills.forEach(originalBill => {
+            combinedBill.originalBills.forEach(async originalBill => {
               const res = result[originalBill._id]
               res.debitOperation = debitOperation
 
-              this.addBillToOperation(originalBill, debitOperation)
+              await this.addBillToOperation(originalBill, debitOperation)
```
ptbrowne: No need for await here, updateAttributes only does synchronous operations and returned Promise.resolve not to change from the previous code. We can remove the Promise.resolve from updateAttributes and remove the async/await here.

packages/cozy-konnector-libs/libs/linkBankOperations.js
```diff
@@ -208,7 +208,9 @@ class Linker {
               const res = result[originalBill._id]
               res.debitOperation = debitOperation
 
-              await this.addBillToOperation(originalBill, debitOperation)
+              if (res.creditOperation && res.debitOperation) {
+                await this.addReimbursementToOperation(originalBill, debitOperation, res.creditOperation)
```
ptbrowne: No need for await here (see above)

Issue #169, opened 2018-04-06
cozy-konnector-libs must be build


kosssi:
Add a build on cozy-konnector-libs to prevent all connectors to install a new dependency for the spread function.

TODO:
- [x] Add build

PR #170, opened 2018-04-06
✨ feat: ship transpiled sources

ptbrowne:
Will fix parsing error when the parser does not support new ES syntaxes.

https://travis-ci.org/konnectors/cozy-konnector-trainline/builds/362851149#L568

- [x] Moved non transpiled sources to `src/`
- [x] Added prepublishOnly task to transpile automatically into `dist/` when `lerna publish`ing
- [x] Changed entrypoint to dist/index.js


kosssi:
Fix https://github.com/konnectors/libs/issues/169

PR #173, opened 2018-04-09
Use connector slug as account_type in yarn dev

### LINKED ISSUE #80: Dev account creation
The accountType should be the name of the konnector because when we want to delete the account, the stack tries to find the konnector associated with the account via accountType.

doubleface:
Fixes #80


packages/cozy-jobs-cli/src/manifest.js
```diff
@@ -14,5 +14,12 @@ module.exports = {
...
 
     return scopes
+  },
+  getSlug (manifestPath) {
+    return getManifest(manifestPath).slug
   }
 }
+
+function getManifest (manifestPath) {
+  return JSON.parse(fs.readFileSync(manifestPath))
```
drazik: Are we 100% sure that the file will exist, so we don't add error handling ?

PR #175, opened 2018-04-09
Update repo paths

sebn:


sebn:
I assume all subpackages are under the MIT license too, or am I wrong?

PR #183, opened 2018-04-10
fix: cozy-logger : check JSON.stringify errors in production

doubleface:
Example, when we try to log an object with circular reference, we cannot JSON.stringify it (ex : an entry with a `filestream` attribute), we get an exception.

We should not get an exception for a logging feature.


kosssi:
il manque un descriptif pour que l'on comprenne l'erreur que tu veux corriger ;)

doubleface:
Avec la description

doubleface:
@ptbrowne Now I use json-stringify-safe but still with error catching for other cases.

PR #188, opened 2018-04-13
fix: Ensure root yarn lint script can actually be run

### LINKED ISSUE #186: ESLint monorepo setup issue
```
$ git clone git@github.com:konnectors/libs.git
$ cd libs/
$ yarn
$ yarn bootstrap
$ yarn lint
sh: eslint : commande introuvable
$ yarn add --dev eslint
$ yarn lint
Cannot find module 'eslint-config-cozy-app'
$ yarn add --dev eslint-config-cozy-app
$ yarn lint
ESLint couldn't find the plugin "eslint-plugin-prettier".
$ yarn add --dev eslint-plugin-prettier
$ yarn lint
... (runs forever)
^C
$ node --version
v8.11.1
$ yarn --version
1.5.1
```

Is it normal that the `yarn lint` script runs `eslint` on the whole monorepo (`.`)?

sebn:
Fixes #186

Although the script is actually not run on CI.
If it did, it would currently reports 700+ issues.


sebn:
Question: what do we want to do in the long term? Use eslint+prettier for all subpackages (eventually replacing standard)?

sebn:
Oops, reverted my mistake.

ptbrowne:
The goal would be to use a common config for all konnector, something akin to `eslint-config-cozy-app`. I think it is OK to run it on the whole repo but we should add packages/*/node_modules to the eslint ignore.

sebn:
I believe `packages/*/node_modules` are already ignored.

Merging since now my editor can at least lint new files.
We can run something like `prettier --fix` on existing code and merge the result in another PR.

PR #191, opened 2018-04-13
feat: Introduce public helper function normalizeFilename()

sebn:


sebn:
Do you think:

- we should allow spaces? (maybe only once at a time?)
- we should apply this by default to `saveFiles()` or `saveBills()` arguments?

packages/cozy-konnector-libs/src/libs/normalizeFilename.js
```diff
@@ -0,0 +1,42 @@
...
+ * used in a filename with underscores.
+ *
+ * Replaced characters include:
+ *
+ * - Those forbidden on one or many popular OS or filesystem: `<>:"/\|?*`
+ * - Those forbidden by the cozy-stack `\0`, `\r` and `\n`
+ * - Spaces and tabs (while those should work, preventing them may magically
+ *   fix issues with other softwares)
+ *
+ * A warning will be logged in case the result doesn't make sense.
```
sebn: Oops, should be `An exception will be thrown in case there is not any filename-compatible character in the given string`.

I wonder whether we could enforce some kind of default name, but sounds convoluted.

sebn: I assume it's the responsibility of the developer to provide enough input to generate a valid filename... Makes sense?

sebn:
- [ ] I also forgot to update the documentation

doubleface:
> we should allow spaces? (maybe only once at a time?)

I don't see a reason to prohibit them

> we should apply this by default to saveFiles() or saveBills() arguments?

It would be good for new connecteur. But for old ones, it depends if all those chars are a cause of bugs.

If it is not the case, we would cause duplicate files for existing connectors since we do not have file migration features yet.

sebn:
- [ ] Allow spaces and use them instead of underscores whenever possible

PR #192, opened 2018-04-13
feat: Introduce mkdirp() function

sebn:


sebn:
Would be useful for connectors that need to create subdirs on the Cozy (e.g. Pajemploi).

sebn:
- [ ] Update documentation (not sure how either, see #191)

doubleface:
For the doc generation, see #194

sebn:
Rebased & generated API doc. I added a trailing slash example but struggled and didn't add any for null path components.

:ship: ?

sebn:
(we may have conflicts with #191 though since both functions will be inserted at the same place in the API documentation...)

PR #214, opened 2018-05-05
WIP feat: add saveBankingDocuments function to cozy-konnector-libs

### LINKED ISSUE #79: Liaison entre les factures et les fichiers
# Problème

Lorsque l'on supprime une facture coté cozy-drive ou cozy-desktop il n'y a pas de suppression du lien dans l'interface de cozy-bank. Ce qui nous retourne une erreur puisque l'on demande à cozy-drive de nous afficher un fichier qui n'existe plus.

# Attendu

Ne plus voir afficher le lien coté cozy-bank après avoir supprimé son fichier coté cozy-drive ou cozy-desktop.

# Explication

Actuellement on fait un lien entre les factures et les fichiers directement dans un champ de la facture. [`bill.invoice`](https://github.com/cozy/cozy-konnector-libs/blob/b9290d10887803c9aa6f61940312401f6782298e/libs/saveBills.js#L31).

Il faudrait utiliser [referenced_by que la stack à mis en place](https://github.com/cozy/cozy-stack/blob/master/docs/references-docs-in-vfs.md).

Les différents impactes que ça peut avoir coté cozy-bank:
- l'application va devoir ajouter le doctype `io.cozy.files` dans ses autorisations.
- modifier l'import des factures

doubleface:
The goal of this PR is to replace saveBills with saveBankingDocuments which will be able to handle more doctypes than io.cozy.bills
and also handles link between documents and bank operations using relationships

Then this also fixes #79

The doctype stays `io.cozy.bills` by default and an option has been added to specify the doctype :

```js
saveDocuments(documents, fields, {
  doctype: 'io.cozy.payslips'
})
```

  At the moment :

  - io.cozy.bills
  - io.cozy.payslips

  These doctype have some attributes in common :

  - amount
  - date
  - vendor
  - $relationships.data.document

  These attributes will make it easier (hopefully) for the cozy-bank application to link theses
  documents to bank operations.

The cozy bank application will have to do some adaptions to be able to display the documents created with this function


packages/cozy-konnector-libs/src/libs/saveBankingDocuments.js
```diff
@@ -1,36 +1,40 @@
...
+ * You need the full permission on the doctype, full permission on `io.cozy.files` and also
+ * full permission on `io.cozy.bank.operations` in your manifest, to be able to use this function.
  *
  * Parameters:
  *
  * - `documents` is an array of objects with any attributes :
  * - `fields` (object) this is the first parameter given to BaseKonnector's constructor
  * - `options` is passed directly to `saveFiles`, `hydrateAndFilter`, `addData` and `linkBankOperations`.
+ *      - `doctype` option has `io.cozy.bills`
+ *      - `noBanking` option deactivates linkBankOperations. Default value is false.
```
sebn: Since we may want to prevent linking with bank operations, doesn't this mean there will be other possible uses of these money-related documents? Then what about `saveFinancialDocuments()` as a name (with *financial* as in *money-related*)?

packages/cozy-konnector-libs/src/libs/saveBankingDocuments.js
```diff
@@ -1,36 +1,40 @@
...
- * common case: bills.
- * Will create `io.cozy.bills` objects. The default deduplication keys are `['date', 'amount', 'vendor']`.
- * You need the full permission on `io.cozy.bills`, full permission on `io.cozy.files` and also
- * full permission on `io.cozy.bank.operations` in your manifest, to be able to * use this function.
+ * Combines the features of `saveFiles`, `filterData`, `addData` and `linkBankOperations` for any
+ * doctype.
+ * Will create `io.cozy.bills` documents by default or any specified doctype.
+ * The default deduplication keys are `['date', 'amount', 'vendor']`.
+ * You need the full permission on the doctype, full permission on `io.cozy.files` and also
+ * full permission on `io.cozy.bank.operations` in your manifest, to be able to use this function.
```
sebn: *... to be able to use this function (unless option `noBanking: true` is set, see below).*?

Or will the permission still be needed anyway?

packages/cozy-konnector-libs/src/libs/saveBankingDocuments.js
```diff
@@ -1,36 +1,40 @@
...
- * const { BaseKonnector, saveBills } = require('cozy-konnector-libs')
+ * const { BaseKonnector, saveBankingDocuments } = require('cozy-konnector-libs')
  *
  * module.exports = new BaseKonnector(function fetch (fields) {
  *   const documents = []
  *   // some code which fills documents
- *   return saveBills(documents, fields, {
- *     identifiers: ['vendorj']
+ *   return saveBankingDocuments(documents, fields, {
+ *     identifiers: ['vendorj'],
```
sebn: `s/vendorj/vendor/`? Or does it have some special meaning?

packages/cozy-konnector-libs/src/libs/saveBankingDocuments.js
```diff
@@ -41,20 +45,32 @@ module.exports = (entries, fields, options = {}) => {
...
+  options.doctype = options.doctype || DEFAULT_DOCTYPE
+
   // Deduplicate on this keys
   options.keys = options.keys || ['date', 'amount', 'vendor']
 
   options.postProcess = function (entry) {
     if (entry.fileDocument) {
-      entry.invoice = `io.cozy.files:${entry.fileDocument._id}`
+      entry.$relationships = entry.$relationships || {}
+      entry.$relationships.document = {
```
sebn: Is `$relationships` a cheerio object? Then couldn't it happen that such a common name as `document` would make its way into cheerio's API?

aenario: `$relationships` is not a cheerio object. It is a cozy-special fields for relationships (see https://github.com/cozy/cozy-doctypes/pull/38)

packages/cozy-konnector-libs/src/libs/saveBankingDocuments.js
doubleface: That is right, I added the noBanking option later.

Then the name of the function should be saveDocuments since we may want to use it without banking, and since the banking part may be extracted from the connector anyway.

Issue #146, opened 2018-03-16
Use yarn workspaces


ptbrowne:
I was not aware of `yarn workspaces` when I split the repos.

As stated here : https://yarnpkg.com/lang/en/docs/workspaces/, lerna and yarn workspaces work beautifully together (jest is using this), and more efficiently : 

* `yarn` does its jobs of maintaining node_modules and symlinking
* `lerna` still helps for publishing modules

I guess it should only be a matter of adding `"useWorkspaces": true` to `lerna.json`, adding `workspaces: "packages/*"` into the main `package.json` and check that it works.

PR #218, opened 2018-05-06
feat: now use yarn workspaces

doubleface:
Fixes : #146


aenario:
Let's be careful that all packages are correct on next release !

PR #219, opened 2018-05-06
print in console the url of the oauth modal

Benibur:
If the script is run in a headless virtual machine, then the autorization oauth modal won't open.
The solution is to manually open a browser with the url loged in the console.


doubleface:
Thanks @benibur ! This may have lost some people indeed!

PR #221, opened 2018-05-08
fix: now the yarn dev command will work again with remote cozies

doubleface:
Some explanations : 

At the moment, the cozy-run-dev command does not work when you define `COZY_URL` in `konnector-dev-config.json` file to something else than `http://cozy.tools:8080`. This is caused by the fact that passwords are encrypted and the development connector has no right to read it.

To fix this, cozy-run-dev won't create an account on the cozy anymore but will modify the BaseKonnector dynamically to give credentials directly to the connector.


drazik:
Do you think this PR could be merged soon ? It seems it can fix an issue I have while testing a connector in dev mode.

doubleface:
@LucsT You can change the url in konnector-dev-config.json and run yarn dev

PR #245, opened 2018-05-23
Add html2pdf

aenario:


aenario:
WIP: not exported from the package

PR #250, opened 2018-05-25
Ran yarn lint

doubleface:


CODEOWNERS
```diff
@@ -1,2 +1,2 @@
 # General code owners
-*      @doubleface @aenario @sebn @LucsT
+*      @doubleface @aenario @sebn @LucsT @enguerran
```
enguerran: you can use the team @konnectors/maintainers instead of a list of members hard to maintain. Cf. https://github.com/konnectors/cozy-konnector-template/issues/92

PR #253, opened 2018-05-30
feat: Automatically currency to get iso code or set 'EUR' by default

LucsT:
This PR need to be discuss with doubleface before merge.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/saveBills.js
+++ b/packages/cozy-konnector-libs/src/libs/saveBills.js
@@ -50,6 +50,7 @@ module.exports = (entries, fields, options = {}) => {
   options.keys = options.keys || ['date', 'amount', 'vendor']
 
   options.postProcess = function(entry) {
+    entry.currency = convertCurrency(entry.currency)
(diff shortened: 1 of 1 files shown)


LucsT:
@doubleface 
Change done and tested.
Now set EUR by default if currency not set and return currency if the sign is not know

packages/cozy-konnector-libs/src/libs/saveBills.js
```diff
@@ -56,6 +56,9 @@ module.exports = (entries, fields, options = {}) => {
     delete entry.fileDocument
     return entry
   }
+  entries.map(entry => {
```
aenario: :warning: This will not work in konnectors where we pass an array of promise (like https://github.com/konnectors/auchandrive/blob/3b9d41d3e2dd0d2804bfcfe442a57bdb5e928b0d/src/index.js#L36)

Put it in `postProcess` instead.

LucsT:
@doubleface @aenario PR updated in postProcess

LucsT:
Tested in a connector with yarn link

PR #255, opened 2018-06-01
refactor: now handle rootPath creation in standalone command script and not in stub

doubleface:


doubleface:
@aenario, since this stub is always called in standalone mode, which always defines COZY_FIELDS, the default value is not needed. I remove it.

packages/cozy-konnector-libs/package.json
```diff
@@ -47,6 +47,7 @@
     "uuid": "^3.1.0"
   },
   "scripts": {
+    "build": "babel src --out-dir dist",
```
enguerran: Why do you duplicate `transpile` script?

doubleface: This is an alias because I prefer yarn build.

I should probably do : "build" : `"npm run transpile"`

PR #265, opened 2018-06-07
feat: activate linting before publish

doubleface:
+ fixed linting issues

### DIFF
--- a/packages/cozy-jobs-cli/src/cozy-authenticate.js
+++ b/packages/cozy-jobs-cli/src/cozy-authenticate.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+/* eslint no-console: off */
 const http = require('http')
 const fs = require('fs')
 const log = require('cozy-logger').namespace('cozy-authenticate')
--- a/packages/cozy-konnector-libs/scripts/visualizer/index.js
+++ b/packages/cozy-konnector-libs/scripts/visualizer/index.js
@@ -24,6 +24,7 @@ const generate = async options => {
 }
 
 process.on('unhandledRejection', x => {
+  // eslint-disable-next-line no-console
   console.warn(x)
 })
 
@@ -43,7 +44,6 @@ const parser = spec => obj => {
   return res
 }
 
-const int = x => parseInt(x, 10)
 const float = x => parseFloat(x, 10)
 const htmlFormBoolean = x => x === 'on'
 const commaSeparatedArray = x => x.split(/\s*,\s*/)
@@ -66,5 +66,5 @@ app.get('/', (req, res) => {
   const index = fs.readFileSync(require.resolve('./index.html'))
   res.send(index.toString())
 })
-
+// eslint-disable-next-line no-console
 app.listen(3000, () => console.log('Visualizer server running'))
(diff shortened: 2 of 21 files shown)


packages/cozy-jobs-cli/src/cozy-authenticate.js
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+/* eslint no-console: off */
```
enguerran: I think we must use the logger

Or put this rule in the global config file

packages/cozy-konnector-libs/scripts/visualizer/index.js
```diff
@@ -24,6 +24,7 @@ const generate = async options => {
 }
 
 process.on('unhandledRejection', x => {
+  // eslint-disable-next-line no-console
```
enguerran: don't like this rule for one line

if we use console.log, just let's put the rule in the config file

packages/cozy-konnector-libs/src/libs/hydrateAndFilter.js
```diff
@@ -133,7 +133,7 @@ const hydrateAndFilter = (documents, doctype, options = {}) => {
   }
 
   const defaultShouldSave = () => true
-  const defaultShouldUpdate = existing => false
+  const defaultShouldUpdate = existing => false // eslint-disable-line no-unused-vars
```
enguerran: mmmhhh, thank you lord eslint!
But it seems that the variable should be a function that always returns `false' regardless of the given parameter.

So why do we have a parameter?

packages/cozy-konnector-libs/src/libs/linkBankOperations.js
```diff
@@ -21,6 +21,7 @@ const flatten = require('lodash/flatten')
 const sumBy = require('lodash/sumBy')
 const geco = require('geco')
 const moment = require('moment')
+const cozyClient = require('./cozyclient')
```
enguerran: I think we could use the variable stored in `global`, don't we?

packages/cozy-logger/src/index.js
```diff
@@ -8,7 +8,7 @@ let level = LOG_LEVEL || 'debug'
 const format = env2formats[env]
 const filters = [filterLevel, filterSecrets]
 
-const filterOut = function(level, type, message, label, namespace) {
+const filterOut = function() {
```
enguerran: That's a change! 5 useless parameters removed! ;-)

packages/cozy-jobs-cli/src/cozy-authenticate.js
doubleface: In this case, console.log is used to display a message to the user of the cozy-run-dev cli. I think this is justified.

packages/cozy-konnector-libs/scripts/visualizer/index.js
doubleface: I actually want to enforce this rule in cozy-konnector-libs. But this file is a script to generate a visualization of links between doctypes and is not run by the stack, then we don't need to use log here.

packages/cozy-konnector-libs/src/libs/hydrateAndFilter.js
doubleface: I chose to still display this parameter to an eventual user of the shouldUpdate option that this parameter is available.

packages/cozy-konnector-libs/src/libs/linkBankOperations.js
doubleface: I don't know, I don't to use globals. I have no better argument

PR #271, opened 2018-06-17
feat: also pass full response to signin validate function

doubleface:
This allows to validate the authentication with response headers for
example


LucsT:
I approve, but there is a merge conflict to solve.

PR #284, opened 2018-06-27
feat: add some generic error codes

doubleface:


enguerran:
There is a small doc in front of each variable declaration. But that is true it could be difficult to choose the correct error to be the more accurate we can. This is a common problem, in Java, it is hard to throw the correct exception, bad developers throw `RuntimeException` no matter what. But with baby steps incrementation, connectors' error will become more accurate.

enguerran:
Autocompletion or something like that could help developers while they are writing connectors.

PR #285, opened 2018-06-27
feat(cozy-konnector-libs): Clean dist directoring before transpiling

LucsT:
I propose to force the deleting of dist directory data when transpiling.
Because babel only update files, you can keep old file and lead to buggy thing (case with new request.js directory)

What do you think about it ?


enguerran:
Thanks for the work, it is very valuable.
Can you edit your commit and pull-request description to take into account the guidelines? https://github.com/cozy/cozy-guidelines#commit-messages

PR #293, opened 2018-07-04
fix: Export signin as submitForm

### LINKED ISSUE #290: Rename signin to make it more discoverable
I am developing a new [konnector](https://github.com/laedit/cozy-konnector-gazdestrasbourg) for [Gaz de Strasbourg](https://agence-en-ligne.es-gazdestrasbourg.fr/) and since it is heavily based on old ASP.NET form navigation, I use `signin` to submit the form for each navigation.

In order to help everyone realize that `signin` isn't just for authentication, I think that it should be renamed like `submitForm`, or at least exposed under two names.

I can propose a PR if you are ok with the idea.

laedit:
Expose `signin` as `submitForm` to ease its use outside of authentication.

fix #290

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "babel-preset-env": "1.7.0",
     "cross-env": "5.2.0",
     "eslint": "5.0.1",
-    "eslint-config-cozy-app": "0.8.0",
+    "eslint-config-cozy-app": "0.10.0",
     "lerna": "2.11.0",
     "lerna-changelog": "0.8.0"
   },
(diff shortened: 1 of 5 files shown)


enguerran:
Thanks @laedit for your work, I updated the commit message and the Pull-Request title to follow our guidelines you can find here: https://github.com/cozy/cozy-guidelines#commit-messages

doubleface:
@laedit I approved, but can you just change the commit message for something like

```
refactor: submitForm alias for the signin function

fix #290
```

We have some guidelines about this now : https://github.com/cozy/cozy-guidelines#commit-messages

laedit:
Thanks for the links, I have missed it.

PR #294, opened 2018-07-05
feat: Log matching bills' vendor and amount

drazik:
Previously, we logged the subtype of the bill. But every bill doesn't have a subtype, and it seems more informatique to know the vendor and the amount of a bill that its subtype.


doubleface:
Since the matching is done in cozy-bank's service. I think this PR is not needed anymore.

PR #295, opened 2018-07-05
feat: Update visualizer default values

drazik:
Use matching algorithm default values as minDateDelta, maxDateDelta and
amountDelta values to better fit real world case by default


enguerran:
I am [very](https://github.com/konnectors/libs/pull/295#event-1718078632) happy! Thanks for the work and the contribution.

PR #296, opened 2018-07-06
Update group matching

drazik:
Before, only health bills were grouped to try to match them with grouped debits. But there are some vendors (like Numéricable) that also do grouped debits.

So we did two things here:

* Update the properties we group bills on : now we group on date and vendor, instead of date and type
* Add an array of vendor that do grouped debits : this way we don't consume RAM to group bills of vendors that don't do grouped debits

The only thing that seems crappy is that we have a hardcoded array of vendors names that do grouped debits. If you think about a better solution for that, I would be happy to read it ! But since in a near future this code will be put in a cozy-banks' service, it is not so problematic.

Here is the result in the visualizer:
image

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/linkBankOperations.js
+++ b/packages/cozy-konnector-libs/src/libs/linkBankOperations.js
@@ -280,7 +280,7 @@ class Linker {
       moment(bill.originalDate)
         .format()
         .split('T')[0],
-      bill.type
+      bill.vendor
     ])
 
     return Object.values(groups)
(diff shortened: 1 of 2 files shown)


packages/cozy-konnector-libs/src/libs/linkBankOperations.js
```diff
@@ -36,6 +36,7 @@ class Linker {
   constructor(cozyClient) {
     this.cozyClient = cozyClient
     this.toUpdate = []
+    this.groupVendors = ['Numéricable']
```
doubleface: Free also does this kind of grouping I think

drazik:
I try to think about memory usage of the algorithm, so we don't have the same problem as before on production. I don't know how to test the memory usage of this on multiple cases. @doubleface do you see something we can do here ?

drazik:
I just added a little performance improvement. But I don't see a reason why it would consume too much memory in prod, so I'm merging.

PR #306, opened 2018-07-17
Configure eslint so that it runs on visualizer files

ptbrowne:
Previously, eslint had been configured to ignore all visualizer files
since there was a big minified files (viz.js) that took a long time
to lint. I added .min.js so that we can add the .min.js to the
eslintignore file.

I also ran eslint on two files.


ptbrowne:
How so ?

enguerran:
You can squash every commits related to eslint configuration and its application in code source formats in order to have one unitary commit and no commits dependent to others and their order of application.

ptbrowne:
Squashed and merged in https://github.com/konnectors/libs/commit/ee595b2cbdba45cc17e4d7f8cd4aa668b6808a73

PR #308, opened 2018-07-18
tests: Add real test cases

ptbrowne:
In this PR, I refactor the test cases for bank operation <-> bill for the
tests to be written declaratively and I add real test cases for Trainline,
Malakoff and Harmonie.

Declarative tests are more readable and are the first step to have only
text files describing the result, the goal being for the PO to be able
to understand the test cases.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/linkBankOperations.spec.js
+++ b/packages/cozy-konnector-libs/src/libs/linkBankOperations.spec.js
@@ -343,6 +343,16 @@ describe('linker', () => {
           })
         })
     })
+
+    it('should not link twice', async () => {
+      const test = tests[0]
+      const options = { ...defaultOptions, ...test.options }
+      expect(operationsById.medecin.reimbursements).toBe(undefined)
+      await linker.linkBillsToOperations(test.bills, options)
+      expect(operationsById.medecin.reimbursements.length).toBe(1)
+      await linker.linkBillsToOperations(test.bills, options)
+      expect(operationsById.medecin.reimbursements.length).toBe(1)
+    })
   })
 
   describe('linking with combinations', () => {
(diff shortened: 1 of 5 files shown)


packages/cozy-konnector-libs/src/libs/linkBankOperations.spec.js
```diff
@@ -243,109 +273,103 @@ describe('linker', () => {
...
-      ]
-      const options = { ...defaultOptions, identifiers: ['SFR'] }
-      return linker
-        .linkBillsToOperations(noHealthBills, options)
-        .then(result => {
-          expect(result).toMatchObject({
-            b2: { debitOperation: operationsById.small_sfr }
+        if (test.dbOperations) {
+          test.dbOperations.forEach(op => operations.push(op))
+          test.dbOperations.forEach(op => {
```
drazik: It seems you can merge the two `forEach` here

packages/cozy-konnector-libs/src/libs/linkBankOperations.spec.js
```diff
@@ -270,6 +270,56 @@ describe('linker', () => {
...
+        },
+        bills: [
+          'b1 | 5.9  |  | 45 | 09-01-2018 | 09-01-2018 | true | Ameli    | health_costs',
+          'b2 | 13.8 |  | 45 | 15-01-2018 | 15-01-2018 | true | Malakoff | health_costs'
+        ],
+        dbOperations: [
+          'malakoff | 15-01-2018 | Malakoff Mederic Pre | 13.8 | 400610',
+          'cpam     | 12-01-2018 | Cpam de Paris        | 5.9  | 400610',
+          'docteur  | 09-01-2018 | Docteur Konqui       | -45  | 400610'
+        result: () => ({}),
```
drazik: I don't understand what this function does here. There's no comma between the previous string and the function :thinking:

ptbrowne: The testing bit is in `operations` here. The comma must have been lost during the various rebases I did to cleanly divise the commits. It is rewritten in the succeeding commits.

ptbrowne:
> I generally don't like declarative tests because I find it harder to read when I want to know what is really tested when the tests fail

Understood ! I will keep this in mind if the test fails and try to improve if I find it hard to understand what happens. I think it is possible to have the best of both world but it is necessary to keep that in mind. Thanks !

PR #310, opened 2018-07-19
tests: more tests

ptbrowne:
Added tests to check for 

- Reimbursements being added to an operation
- Reimbursements cannot be added to an operation when it is "full"
- Matching favoring groupAmount if present


kosssi:
C'est vraiment super lisible comme test 👍 

mini retour tu as pas mis les espaces pour correspondre au titre [ici](https://github.com/konnectors/libs/pull/310/files#diff-b4c035e16ede8aa57e799e09d6a2e762R365) et [ici](https://github.com/konnectors/libs/pull/310/files#diff-b4c035e16ede8aa57e799e09d6a2e762R389)

ptbrowne:
The failing test is a timeout of request.spec.js.

PR #322, opened 2018-08-07
fix(hydrateAndFilter): fetch all db items

doubleface:
The hydrateAndFilter function did not fetch all db items because of 100 limit. This causes some items to be duplicated

fixes https://github.com/konnectors/libs/issues/323

### DIFF
--- a/packages/cozy-konnector-libs/docs/api.md
+++ b/packages/cozy-konnector-libs/docs/api.md
@@ -231,6 +231,58 @@ use this function.</p>
 is already saved in the cozy</li>
 </ul>
 </dd>
+<dt><a href="#module_filterData">filterData</a></dt>
+<dd><p>This module proposes some small utils regarding connectors</p>
+<p>Parameters:</p>
+<ul>
+<li><code>documents</code>: an array of objects corresponding to the data you want to save in the cozy</li>
+<li><code>doctype</code> (string): the doctype where you want to save data (ex: &#39;io.cozy.bills&#39;)</li>
+<li><code>options</code> :<ul>
+<li><code>keys</code> (array) : List of keys used to check that two items are the same. By default it is set to `[&#39;id&#39;]&#39;.</li>
+<li><code>index</code> (optionnal) : Return value returned by <code>cozy.data.defineIndex</code>, the default will correspond to all documents of the selected doctype.</li>
+<li><code>selector</code> (optionnal object) : Mango request to get records. Default is built from the keys <code>{selector: {_id: {&quot;$gt&quot;: null}}}</code> to get all the records.</li>
+</ul>
+</li>
+</ul>
+<pre><code class="lang-javascript">const documents = [
+  {
+    name: &#39;toto&#39;,
+    height: 1.8
+  },
+  {
+    name: &#39;titi&#39;,
+    height: 1.7
+  }
+]
+
+return filterData(documents, &#39;io.cozy.height&#39;, {
+  keys: [&#39;name&#39;]
+}).then(filteredDocuments =&gt; addData(filteredDocuments, &#39;io.cozy.height&#39;))
+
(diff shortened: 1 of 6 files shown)


packages/cozy-konnector-libs/src/libs/hydrateAndFilter.js
```diff
@@ -107,7 +107,18 @@ const hydrateAndFilter = (documents = [], doctype, options = {}) => {
...
-    return cozy.data.query(index, { selector })
+    const result = []
+    while (true) {
+      log('debug', `Fetching next db items ${result.length}`)
+      const resp = await cozy.data.query(index, {
+        selector,
+        wholeResponse: true,
+        skip: result.length
+      })
+      result.push.apply(result, resp.docs)
```
enguerran: it might be easier to read using [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax):
```js
result.push(...resp.docs);
```

packages/cozy-konnector-libs/src/libs/hydrateAndFilter.js
```diff
@@ -107,7 +107,18 @@ const hydrateAndFilter = (documents = [], doctype, options = {}) => {
...
+      if (!resp.next) break
```
enguerran: Naive question: why don't you loop on this conditional expression?

```js
while(resp.next) {
}
```

packages/cozy-konnector-libs/src/libs/hydrateAndFilter.spec.js
```diff
@@ -26,10 +26,12 @@ describe('hydrate and filter', () => {
...
-      asyncResolve([
-        { _id: 1, name: 'Marge', _rev: 2 },
-        { _id: 2, name: 'Homer', _rev: 3 }
-      ])
+      asyncResolve({
+        docs: [
+          { _id: 1, name: 'Marge', _rev: 2 },
+          { _id: 2, name: 'Homer', _rev: 3 }
+        ]
+      })
```
enguerran: How does this test cover the pagination?

packages/cozy-konnector-libs/src/libs/hydrateAndFilter.js
doubleface: Right, fixed this before seeing your comment

packages/cozy-konnector-libs/src/libs/hydrateAndFilter.spec.js
doubleface: This test does not cover the pagination, it is just a fix to allow this test to work

packages/cozy-konnector-libs/docs/api.md
```diff
@@ -231,6 +231,58 @@ use this function.</p>
...
+<p>Parameters:</p>
+<ul>
+<li><code>doctype</code> (string): the doctype from which you want to fetch the data</li>
+<li><code>selector</code> (object): the mango query selector</li>
+<li><code>index</code> (object): (optional) the query selector index. If not defined, the function will
+create it&#39;s own index with the keys specified in the selector</li>
+</ul>
+<pre><code class="lang-javascript">const documents = await queryAll(&#39;io.cozy.bills&#39;, {vendor: &#39;Direct Energie&#39;})
+</code></pre>
+</dd>
```
enguerran: What the point to write html code in a markdown file?

doubleface: This is generated code actually : https://github.com/konnectors/libs/blob/3614911cc293510cd8ae29cb01b42bc789bc2ea3/packages/cozy-konnector-libs/package.json#L44

LucsT:
I'am now! ^^

PR #324, opened 2018-08-08
feat: check for duplicates in saveBills

doubleface:
Also add utils to help with the cleaning


packages/cozy-konnector-libs/src/helpers/cozy-client-js-stub.js
```diff
@@ -56,10 +56,14 @@ module.exports = {
...
       const { selector } = options
       const keys = Object.keys(selector)
-      const result = db
+      let result = db
         .get(doctype)
         .filter(doc => keys.every(key => doc[key]))
         .value()
+
+      if (options.wholeResponse) {
+        result = { docs: result }
```
enguerran: Reusing a variable name is a smell in a function's code that responsibilities are not well established.
```
db
        .get(doctype)
        .filter(doc => keys.every(key => doc[key]))
        .value()
```
may be a heavy processing, and never be used if `option.wholeResponse` is `true`.

A `if/then` seems a better solution here.

PR #337, opened 2018-08-29
WIP: fix: keep bills linked to operations if any

doubleface:
Here the bills are first ordered by the number of linked operations so that the bills with the most of linked
operation are kept. This is a work in progress and not tested yet.

PR #338, opened 2018-08-29
test: Update test of linkBankOperations

LucsT:
Just for advice and structure. @doubleface 
Will start implementation


LucsT:
No need to keep it. 
Merge will done from : https://github.com/konnectors/libs/pull/337

PR #365, opened 2018-11-06
feat: Extend from cozy configuration

y-lohse:
For reference, here is the cozy config : https://github.com/cozy/cozy-libs/blob/master/packages/renovate-config-cozy/package.json#L5-L23

The man change would be that PRs for non-cozy packages would now be made every day **during non office hours**, whereas at the moment they are only made on monday mornings. Would that be a problem?

### DIFF
--- a/packages/renovate-config-cozy-konnector/package.json
+++ b/packages/renovate-config-cozy-konnector/package.json
@@ -10,7 +10,12 @@
       ],
       "minor": {
         "automerge": true
-      }
+      },
+      "packageRules": [
+        {
+          "schedule": "before 5am every monday"
+        }
+      ]
     }
   }
 }
--- a/packages/renovate-config-cozy-konnector/package.json
+++ b/packages/renovate-config-cozy-konnector/package.json
@@ -6,22 +6,11 @@
   "renovate-config": {
     "default": {
       "extends": [
-        "config:base"
+        "cozy"
       ],
-      "timezone": "Europe/Paris",
       "minor": {
         "automerge": true
-      },
-      "packageRules": [
-        {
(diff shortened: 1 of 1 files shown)


packages/renovate-config-cozy-konnector/package.json
```diff
@@ -6,22 +6,11 @@
...
-      },
-      "packageRules": [
-        {
-          "excludePackageNames": [
-            "cozy-konnector-libs",
-            "cozy-app-publish"
-          ],
-          "schedule": "before 5am every monday"
-        }
-      ],
```
y-lohse: The cozy config has almost the same rule, except that:

- it matches all `cozy-*` packages, not just the 2 specified above
- the scheduling happens every day, but still during non office hours

packages/renovate-config-cozy-konnector/package.json
```diff
@@ -6,22 +6,11 @@
...
-      "prHourlyLimit": 0
```
y-lohse: [0 is the fault value](https://renovatebot.com/docs/configuration-options/#prhourlylimit), so I don't think we need this?

doubleface: I suppose you are right

packages/renovate-config-cozy-konnector/package.json
doubleface: For the connectors, every day is two much for 100 connectors. I prefer to keep the previous rule. But I am Ok for everything else.

y-lohse:
Ok no problem, thanks for amending! Feel free to merge this whenever you like :)

PR #375, opened 2018-11-28
feat(saveBills): add processPdf option to get bills data in pdf content

doubleface:


packages/cozy-konnector-libs/src/libs/saveBills.js
```diff
@@ -32,53 +54,65 @@
...
-// Encapsulate the saving of Bills : saves the files, saves the new data, and associate the files
-// to an existing bank operation
-module.exports = (entries, fields, options = {}) => {
+module.exports = async (entries, fields, options = {}) => {
   if (!_.isArray(entries) || entries.length === 0) {
     log('warn', 'saveBills: no bills to save')
     return Promise.resolve()
   }
 
-  checkRequiredAttributes(entries)
```
LucsT: We do not check Attributes anymore ?(as report in personne should not be remove)

doubleface: This is just a move

PR #384, opened 2018-12-12
WIP: feat(libs): Remove bills matching algorithm

drazik:
The algorithm has been ported in [banks service](https://github.com/cozy/cozy-banks/blob/master/src/targets/services/onOperationOrBillCreate.js#L45-L104). This PR goal is to remove the code that we migrated from cozy-konnector-libs to banks.

We don't want to merge this for now because we need to address an issue before stopping matching in connectors: how to configure the algorithm parameters (dates window, amount delta...) from banks' service? We have some plans to do this via auto-descripting bills doctype, but we have not decided anything for now. We need to discuss this more.


drazik:
@doubleface and @LucsT I let you do what you think is best with this PR. It's old and have a lot of conflict but I think that's some work that will be merged someday.

doubleface:
This removal is not relevant anymore

Issue #388, opened 2018-12-23
Docs for API is full of duplicate and messy


ko-dever:
Hello.

[The documentation for API](https://github.com/konnectors/libs/blob/master/packages/cozy-konnector-libs/docs/api.md) is quite messy and full of duplicate (every definition, example and snippets can be found twice in the document, the snippets are in raw text and coloured, ...).

Is this intentional ?

I was about to clean and rearrange the document, but maybe it is auto-generated or whatever.

clochix:
Hello @ko-dever,

Thanks for reporting. I guess some part of this documentation may be auto-generated so I'm pinging the developer.

ptbrowne:
I think it's better now that #389 is merged. Do you confirm @ko-dever  ? Thanks for reporting the issue 👍

ptbrowne:
I think it is added to things that are exported. I am not really sure as it
is JSDoc that adds this :/ Cool that you find it better, docs is always
hard to priorize against other things, but do not  hesitate if you have any
suggestions.

Closing this :)

PR #389, opened 2018-12-24
docs: Do not duplicate info

ptbrowne:
We have duplicate content in the docs since JSdoc puts the docstring content used where there is `@module` both for the table of contents and for the module content.

To document a function that happen to be exported as module, you need to use the alias function. The top description will be used in the table of contents whereas the bottom decription will be used for the documentation when clicking on the link.

```
/** 
 * Short description
 * 
 * @module toto
 */

/** 
 * Long description blablabla
 * @alias module:toto
 */
module.exports = () => {}
```


ptbrowne:
Related: https://github.com/konnectors/libs/issues/388

PR #394, opened 2019-01-08
Make sync field an object indexed by id

Crash--:
We try to follow as can as possible this spec for the next version of our Contacts App. Since 1 contact can be synchronized with multiple sources, we need to change the shape of this object. 

What do you think about this change?

### DIFF
--- a/packages/cozy-konnector-libs/docs/synchronization.md
+++ b/packages/cozy-konnector-libs/docs/synchronization.md
@@ -67,10 +67,16 @@ To store synchronization data, we are using a `sync` attribute in document `meta
 {
   "metadata": {
     "sync": {
-      "id": "7ee401e841c94159addb47f190903139",
-      "konnector": "trainline",
-      "last_sync": "Mon, 12 Feb 2018 16:25:34 GMT"
-    }
+      "7ee401e841c94159addb47f190903139": {
+        "id": "7ee401e841c94159addb47f190903139",
+        "konnector": "trainline",
+        "last_sync": "Mon, 12 Feb 2018 16:25:34 GMT"
+      },
+      "7ee401e841c94159addb47f1909ZZZZ": {
+        "id": "7ee401e841c94159addb47f1909ZZZZZ",
+        "konnector": "icloud",
+        "last_sync": "Mon, 18 Feb 2018 16:25:34 GMT"
+      },
   }
 }
 ```
--- a/packages/cozy-konnector-libs/docs/synchronization.md
+++ b/packages/cozy-konnector-libs/docs/synchronization.md
@@ -70,12 +70,12 @@ To store synchronization data, we are using a `sync` attribute in document `meta
       "7ee401e841c94159addb47f190903139": {
         "id": "7ee401e841c94159addb47f190903139",
         "konnector": "trainline",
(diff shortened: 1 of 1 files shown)


packages/cozy-konnector-libs/docs/synchronization.md
```diff
@@ -67,10 +67,16 @@ To store synchronization data, we are using a `sync` attribute in document `meta
...
   "metadata": {
     "sync": {
-      "id": "7ee401e841c94159addb47f190903139",
-      "konnector": "trainline",
-      "last_sync": "Mon, 12 Feb 2018 16:25:34 GMT"
-    }
+      "7ee401e841c94159addb47f190903139": {
+        "id": "7ee401e841c94159addb47f190903139",
+        "konnector": "trainline",
+        "last_sync": "Mon, 12 Feb 2018 16:25:34 GMT"
```
ptbrowne: Why isn't this a ISO date ?

ptbrowne:
Fine by me but I haven't touched to synchronization much. I'd like to also have the input of @doubleface and @nono .

packages/cozy-konnector-libs/docs/synchronization.md
```diff
@@ -67,10 +67,16 @@ To store synchronization data, we are using a `sync` attribute in document `meta
 {
   "metadata": {
     "sync": {
-      "id": "7ee401e841c94159addb47f190903139",
-      "konnector": "trainline",
-      "last_sync": "Mon, 12 Feb 2018 16:25:34 GMT"
-    }
+      "7ee401e841c94159addb47f190903139": {
+        "id": "7ee401e841c94159addb47f190903139",
```
nono: Why `7ee401e841c94159addb47f190903139` is repeated? First as the key, and then as an id.

packages/cozy-konnector-libs/docs/synchronization.md
Crash--: As the co-writer of this spec, tell me! :D 

Being serious, you'll use the ISO format `2018-10-19T10:58:37.025688+02:00` if @gregorylegarec has no objection ;)

nono:
Do you aim synchronization in only one direction (import from an external service to cozy)?

nono:
I'm also curious to know why we don't use the `importedAt` / `importedFrom` described in https://github.com/cozy/cozy-doctypes/tree/master/docs#document-metadata

packages/cozy-konnector-libs/docs/synchronization.md
nono: OK, if it is more convenient, it's OK for me.

Crash--:
Understood. Let me finish our spec and I'll update this document with all the needed modifications. I just wanted to have an input for the proposed change at the beginning. 

But to answer your question @nono : I can have a contact importedFrom: Google-Konnector, but also synched with iCloud.

nono:
OK, you can still do a `s/metadata/cozyMetadata/`

nono:
Ping @Crash--

Crash--:
@doubleface If I read from https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.contacts.md it seems that sync is into `cozyMetadata`. But here, it's in `metadata`. Do you know the final word?

PR #429, opened 2019-03-13
Add categorization functions

drazik:
This extracts categorization functions from cozy-banks. The reason is because we want to execute the categorization model in banking connectors instead of in a bank's service. This will avoid the debounce time (3mn) during which the categorization model have not been executed and the categories shown to the user in Banks is not optimal and subject to sudden changes that may not be understandable for the user.


doubleface:
@drazik Is it expected to be used outside BI connectors scope (other bank connectors) ? In this case I suppose it should be documented.
-

drazik:
Yes I'm going to write a documentation for contributors that want to use categorization in their banking connectors. Would it be OK for you if we merge this PR as is, and I open another one quickly for documentation?

PR #430, opened 2019-03-15
docs(konnector-libs): Add doc for categorization

drazik:
I only added `src/libs/categoriation/index.js` to API doc because other files are internals and should not be used directly by the lib users.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/categorization/index.js
+++ b/packages/cozy-konnector-libs/src/libs/categorization/index.js
@@ -1,8 +1,41 @@
+/**
+ * Bank transactions categorization
+ * @module categorization
+ */
+
 const { tokenizer } = require('./helpers')
 const { globalModel } = require('./globalModel')
 const { localModel } = require('./localModel')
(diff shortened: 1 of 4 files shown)


packages/cozy-konnector-libs/src/libs/categorization/index.js
```diff
@@ -1,8 +1,41 @@
...
+
 const { tokenizer } = require('./helpers')
 const { globalModel } = require('./globalModel')
 const { localModel } = require('./localModel')
 
-const categorize = async transactions => {
+/**
+ * Apply both global and local categorization models to an array of transactions.
+ *
+ * The global model is a model specific to hosted Cozy instances. It is not available for self-hosted instances. It will just do nothing in that case.
```
ptbrowne: Please wrap the line :)

ptbrowne:
> because other files are internals and should not be used directly by the lib users.

Another solution is to mark documented functions as `@private`.

LGTM 👍

PR #476, opened 2019-05-07
test(categorization): Add dashboard tests

drazik:
This dashboard was previously living inside Banks. But since we moved
the categorization logic in cozy-konnector-libs to be able to use it in
connectors, it makes sense to also move the related testing dashboard.


francoisWeber:
When launching the test, plenty of `console.log` are displayed : would it be possible to mute them ? 
I think we used a 
```js
window.console.log = jest.fn();
window.console.warn = jest.fn();
```
before

drazik:
Actually you can hide console output with the [`--silent` Jest flag](https://jestjs.io/docs/en/cli#silent). Would it work for you?

francoisWeber:
It kind of work : now i see an error that was hidden before : 
`Cannot find module 'csv-writer' from 'services-fixtures.spec.js'` : the package `csv-writer` might not have been added to the dev-dependencies ?

drazik:
Oh yeah I forgot this one it seems. I will add it.

drazik:
@francoisWeber here it is.

francoisWeber:
Thanks !

PR #487, opened 2019-05-20
feat(Categorization): Set global model dashboard

francoisWeber:
Added a snapshot test to measure accuracy of the global model without local corrections. Roughly the same code as `categorization/dashboard/services-fixtures.spec.js `


francoisWeber:
> It seems some parts could be extracted in helpers so they are not repeated, no? Mostly the formatting and output parts.

Yep, will do so. Thanks for the advice

francoisWeber:
> > It seems some parts could be extracted in helpers so they are not repeated, no? Mostly the formatting and output parts.
> 
> Yep, will do so. Thanks for the advice

Should be good now, @drazik

PR #497, opened 2019-06-14
feat: Add saveIdentity in cozy-konnector-libs

LucsT:
As describe, final version.
include documentation modifying
include updateOrCreate modifying (for more possibilities concerning matching attributes)


packages/cozy-konnector-libs/docs/api.md
```diff
@@ -451,6 +454,36 @@ You need the full permission on `io.cozy.files` in your manifest to use this fun
 ```javascript
 await saveFiles([{fileurl: 'https://...', filename: 'bill1.pdf'}], fields)
 ```
+<a name="module_saveIdentity"></a>
+
+## saveIdentity
+Helper to set or merge identities
```
doubleface: A link to https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.identities.md would be better for documentation

PR #515, opened 2019-07-15
refactor: Separate log formats

ptbrowne:
This way it is possible to instruct webpack not to include the dev-format.
This is important since dev-format imports chalk.


doubleface:
cozy-logger is now in cozy-libs

PR #519, opened 2019-07-18
feat (saveBills): add matchingCriterias handling to saveBills

doubleface:
- if specified as an attribute of bills, this attribute is recorded
- else, if the manifest of the connector has any 'banksTransactionRegExp' attribute, add
  this attribute as `matchingCriterias.labelRegex` to the bill
- existing bills will be updated with matchingCriterias if any
- else do nothing

The different possible matchingCriterias as documented in https://github.com/cozy/cozy-doctypes/pull/103

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/saveBills.spec.js
+++ b/packages/cozy-konnector-libs/src/libs/saveBills.spec.js
@@ -1,0 +1,78 @@
+jest.mock('./saveFiles')
+jest.mock('./hydrateAndFilter')
+jest.mock('./addData')
+const addData = require('./addData')
+const hydrateAndFilter = require('./hydrateAndFilter')
+const saveFiles = require('./saveFiles')
+const saveBills = require('./saveBills')
+const manifest = require('./manifest')
+const logger = require('cozy-logger')
+const asyncResolve = val => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve(val)
+    }, 1)
+  })
+}
+
+logger.setLevel('critical')
+
+beforeEach(async function() {
+  saveFiles.mockImplementation(entries => {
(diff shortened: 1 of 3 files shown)


packages/cozy-konnector-libs/src/libs/saveBills.spec.js
```diff
@@ -0,0 +1,78 @@
...
+            vendor: 'vendor'
+          }
+        ],
+        {
+          linkBankOperations: false
+        }
+      )
+    } catch (err) {
+      expect(err.message).toEqual(
+        'saveBills: an entry has a amount which does not respect isNumber'
```
drazik:
```suggestion
        'saveBills: an entry has an amount which does not respect isNumber'
```

packages/cozy-konnector-libs/src/libs/saveBills.spec.js
```diff
@@ -0,0 +1,78 @@
...
+beforeEach(async function() {
+  saveFiles.mockImplementation(entries => {
+    return asyncResolve(entries.map(doc => ({ ...doc, fileDocument: true })))
+  })
+  hydrateAndFilter.mockImplementation(entries => {
+    return asyncResolve(entries)
+  })
+  addData.mockImplementation(entries => {
+    return asyncResolve(entries)
+  })
```
drazik: I don't know if I really understand your needs here, but I'm wondering if you can't use [`mockResolvedValue`](https://jestjs.io/docs/en/mock-function-api#mockfnmockresolvedvaluevalue)?

doubleface: I don't this I can use it here since I want to use the input value as response

drazik: Oh yes OK I see :+1:

PR #537, opened 2019-08-07
feat: add metadata deduplication

doubleface:
This PR will add the possibility to saveFiles (and saveBills) to deduplicate files using data saved in file metadata.

```js
await saveFiles([{fileurl: 'https://toto.com/15ab68', filename: 'file.txt'}], fields, {
  filePrimaryKeys: ['fileurl']
})
```

Here the file `file.txt` will be saved to the destination path with the following metadata : 

```json
{
  "filePrimaryKeys": "https://toto.com/15ab68"
}
```

Some important consequences of this PR :

 - If the filePrimaryKeys option is not used by the connector, nothing changes : the file path will be used as deduplication key
 - files which already exist but do not have metadata will be automatically updated with correct deduplication metadata if the file was not moved or renamed before.

### DIFF
--- a/packages/cozy-konnector-libs/docs/api.md
+++ b/packages/cozy-konnector-libs/docs/api.md
@@ -452,7 +452,7 @@ You need the full permission on `io.cozy.files` in your manifest to use this fun
   + `validateFile` (function) default: do not validate if file is empty or has bad mime type
   + `validateFileContent` (boolean or function) default false. Also check the content of the file to
   recognize the mime type
-  + `filePrimaryKeys` (array of strings). Describe which attributes of files will be taken as primary key for
+  + `fileIdAttributes` (array of strings). Describes which attributes of files will be taken as primary key for
   files to check if they already exist, even if they are moved. If not given, the file path will
   used for deduplication as before.
 
@@ -460,7 +460,7 @@ You need the full permission on `io.cozy.files` in your manifest to use this fun
 **Example**  
 ```javascript
 await saveFiles([{fileurl: 'https://...', filename: 'bill1.pdf'}], fields, {
-   filePrimaryKeys: ['fileurl']
+   fileIdAttributes: ['fileurl']
 })
 ```
 <a name="module_saveIdentity"></a>
--- a/packages/cozy-konnector-libs/docs/api.md
+++ b/packages/cozy-konnector-libs/docs/api.md
@@ -412,7 +412,7 @@ Saves the files given in the fileurl attribute of each entries
 
 You need the full permission on `io.cozy.files` in your manifest to use this function.
 
-- `files` is an array of object with the following possible attributes :
+- `files` is an array of objects with the following possible attributes :
 
   + fileurl: The url of the file (can be a function returning the value). Ignored if `filestream`
   is given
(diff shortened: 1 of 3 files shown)


packages/cozy-konnector-libs/src/libs/saveFiles.spec.js
```diff
@@ -267,3 +273,62 @@ describe('saveFiles', function() {
...
+    it('when the file does not exist, should not return any file', async () => {
+      queryAll.mockReturnValue(asyncResolve([]))
+      cozyClient.files.statByPath.mockReturnValue(asyncReject(false))
+      const result = await getFileIfExists(
+        { filename: 'testfile.txt' },
+        options
+      )
+      expect(result).toBe(false)
+    })
+    it('when the file exists, should return the file', async () => {
```
LucsT: We only test the case with one attribute, it could be interesting to add more valuable tests :
ex:  
 - input an array with 2 attributes for deduplication
 - Check that attributes is correctly processed (trough the 'calculate', and then when passed to 'queryAll')

packages/cozy-konnector-libs/src/libs/saveFiles.js
```diff
@@ -519,3 +644,7 @@ function getValOrFnResult(val, ...args) {
     return val.apply(val, args)
   } else return val
 }
+
+function calculateFileKey(entry, filePrimaryKeys) {
+  return filePrimaryKeys.map(key => get(entry, key)).join('####')
```
LucsT: We joining all attributes here, as discuss, we confirmed that a futur change in a connector code, on the sending of file by differents pieces of code could resume in a different order.

Need a sort to be sure to keep always the same order

LucsT: Sort added in next commit

packages/cozy-konnector-libs/docs/api.md
```diff
@@ -452,11 +452,16 @@ You need the full permission on `io.cozy.files` in your manifest to use this fun
   + `validateFile` (function) default: do not validate if file is empty or has bad mime type
   + `validateFileContent` (boolean or function) default false. Also check the content of the file to
   recognize the mime type
+  + `filePrimaryKeys` (array of strings). Describe which attributes of files will be taken as primary key for
+  files to check if they already exist, even if they are moved. If not given, the file path will
+  used for deduplication as before.
```
ptbrowne: It looks like it is the same thing as `idAttributes`, would you consider renaming the variable ?

https://github.com/cozy/cozy-libs/blob/7a0a26b959af8cf7f9be0608c8e6b4648d927131/packages/cozy-doctypes/README.md#idattributes

packages/cozy-konnector-libs/src/libs/saveFiles.js
```diff
@@ -188,24 +213,102 @@ const removeFile = async function(file) {
...
+    } else {
+      try {
+        const result = await cozy.files.statByPath(
+          getFilePath({ entry, options })
+        )
+        return result
+      } catch (err) {
+        log('debug', err.message)
+        return false
+      }
```
ptbrowne: The try/catch can be extracted to a function `getFileByPathIfExists` since it is the same as above.

packages/cozy-konnector-libs/docs/api.md
doubleface: Good point, but then I would prefer fileIdAttributes since theses options are also given to saveBills (which uses the `keys` option for deduplication) and bills and files do not have the same primary key.

ex: a payment schedule file can be associated to 12 monthly bills

`saveBills` and then `hydrateAndFilter` could use an `idAttributes` alias for `keys` for consistency with `cozy-doctypes`

@edas what do you think

packages/cozy-konnector-libs/src/libs/saveFiles.js
```diff
@@ -188,24 +213,102 @@ const removeFile = async function(file) {
...
+      const result = await cozy.files.statByPath(
+        getFilePath({ entry, options })
+      )
+      return result
+    } catch (err) {
+      log('debug', err.message)
+      return false
+    }
+  } else {
+    const index = await cozy.data.defineIndex('io.cozy.files', [
```
ptbrowne: I would extract all the `else`  to a function.

packages/cozy-konnector-libs/src/libs/saveFiles.js
doubleface: I did a small refactor and used another name bit should be better

packages/cozy-konnector-libs/docs/api.md
edas: Good catch @ptbrowne 
We should probably avoid creating different namings.

👍 with fileIdAttributes. `saveBills` indeed need a specific fileIdAttributes

`saveFile` could allow both idAttributes and fileIdAttributes so it keeps compatibility with both `saveBills` and the generic `idAttributes` naming.

PR #575, opened 2019-08-30
feat: Add notifySuccessfulLogin and deactivateAutoSuccess methods

ptbrowne:
Adds methods useful for konnectors dealing with 2FA.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/BaseKonnector.js
+++ b/packages/cozy-konnector-libs/src/libs/BaseKonnector.js
@@ -280,6 +280,33 @@ class BaseKonnector {
   }
 
   /**
+   * Tells Cozy-Home that we have successfully logged in.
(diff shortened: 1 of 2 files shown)


Crash--:
Does it require some update in Home ? Because in harvest, we also wait 8s before dispatching a loginSuccess https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/connections/triggers.js#L34

Is `HANDLE_LOGIN_SUCCESS` a new event to manage?

ptbrowne:
It's already dealt with in harvest. See https://github.com/cozy/cozy-libs/blob/e94d3f74ab68fc90ee3de9e1f71067691ff7fdc8/packages/cozy-harvest-lib/src/helpers/accounts.js#L35

PR #580, opened 2019-09-04
feat: add autologin handling to the signin function

doubleface:
If called via BaseKonnector though


packages/cozy-konnector-libs/src/libs/BaseKonnector.js
```diff
@@ -216,7 +217,7 @@ class BaseKonnector {
    * Notices that 2FA code is needed and wait for the user to submit it.
    * It uses the account to do the communication with the user.
    *
-   * It 
+   * It
```
ptbrowne: I think we should remove the above line :)

ptbrowne:
Please add a test :)

doubleface:
Here are the unit tests

PR #586, opened 2019-09-06
feat: Overridable main

ptbrowne:
Roles of BankingKonnector methods are clearer.

- Run initializes, calls main, and deals with error/success
- Main is fetch() right now but could be overrided inside konnectors
- Added tests

### DIFF
--- a/packages/cozy-jobs-cli/src/konnector-dev.js
+++ b/packages/cozy-jobs-cli/src/konnector-dev.js
@@ -72,7 +72,7 @@ authenticate({ tokenPath: token, manifestPath: manifest })
     BaseKonnector.prototype.getAccount = async () => {
       return { _id: 'dev-konnector-account-id' }
     }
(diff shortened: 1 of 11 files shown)


packages/cozy-jobs-cli/src/konnector-dev.js
```diff
@@ -59,52 +57,29 @@ if (!manifest && file) {
...
-      return { _id: 'dev-konnector-account-id' }
-    }
-    BaseKonnector.prototype.init = async () => {
-      let rootPath = DEFAULT_ROOT_PATH
-      try {
-        await mkdirp(rootPath)
-      } catch (e) {
+const token =
+  program.token ||
+  DEFAULT_TOKEN_PATH(async () => {
```
ptbrowne: What is going on here ? It seems we are calling a string as a function.

PR #599, opened 2019-10-03
Refactor error handling

ptbrowne:
Instead of binding error handlers at import time, bind them
when BaseKonnector is built, it is cleaner. I don't remember
specifically why I started this refactor, but it was lingering
on my local branches.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/error.js
+++ b/packages/cozy-konnector-libs/src/libs/error.js
@@ -43,10 +43,10 @@ const attachProcessEventHandlers = (prcs = process) => {
 
   return () => {
     attached = false
-    prcs.off('uncaughtException', handleUncaughtException)
-    prcs.off('unhandledRejection', handleUnhandledRejection)
-    prcs.off('SIGTERM', handleSigterm)
-    prcs.off('SIGINT', handleSigint)
+    prcs.removeEventListener('uncaughtException', handleUncaughtException)
+    prcs.removeEventListener('unhandledRejection', handleUnhandledRejection)
+    prcs.removeEventListener('SIGTERM', handleSigterm)
+    prcs.removeEventListener('SIGINT', handleSigint)
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/cozy-konnector-libs/src/libs/error.js
```diff
@@ -1,19 +1,55 @@
...
+/**
+ * Attach event handlers to catch uncaught exceptions/rejections and signals.
+ * Log them as critical and exit the process accordingly.
+ * If the cleanup function has not been called, calling again the function
+ * is a no-op.
+ *
+ * @param  {Process} prcs - Process object, default to current process
+ * @return {Function} When called, removes the signal handlers
+ */
+const attachProcessEventHandlers = (prcs = process) => {
```
doubleface: Why do you pass a prcs parameters when global process is used in event handlers anyway ?

packages/cozy-konnector-libs/src/libs/error.js
```diff
@@ -1,19 +1,55 @@
...
+  }
+  attached = true
+  prcs.on('uncaughtException', handleUncaughtException)
+  prcs.on('unhandledRejection', handleUnhandledRejection)
+  prcs.on('SIGTERM', handleSigterm)
+  prcs.on('SIGINT', handleSigint)
+
+  return () => {
+    attached = false
+    prcs.off('uncaughtException', handleUncaughtException)
```
doubleface: `EventEmitter.off` is marked as added in node v10 (https://nodejs.org/api/events.html#events_emitter_off_eventname_listener) and we still use node v8 in production.

Unit tests worked in node 8 anyway (no idea why) but could you use `removeListener` instead?

packages/cozy-konnector-libs/src/libs/error.js
ptbrowne: For tests.

packages/cozy-konnector-libs/src/libs/error.js
ptbrowne: Done, thanks :)

PR #600, opened 2019-10-07
Add subPath to saveFiles

LucsT:
The goal of this PR is to add to saveFiles.js a subPath option globally and as an attribute of a file.
(And save file in this directory, creating it if needed)


LucsT:
@doubleface 
Here is a first commit seems functionnal in both case from two quick test. Need probably more testing about double slash and things like this.

I begin to write formal tests to add to this commit.

Some in files docs has been began too.

Issue #701, opened 2020-08-17
(jobs-cli/konnector-dev) Token not regenerate after manifest modification


JF-Cozy:
According to https://github.com/cozy/cozy.github.io/blob/dev/src/howTos/dev/services.md cozy-konnector-dev is used to test services. If the app's manifest is modified after the generation of the token, the token is not updated. So we have to delete the token to take into account the changes.

PR #707, opened 2020-09-02
docs: Add token precision for cozy-konnector-dev

JF-Cozy:
Added clarification on using token with cozy-konnector-dev

### DIFF
--- a/packages/cozy-jobs-cli/README.md
+++ b/packages/cozy-jobs-cli/README.md
@@ -20,6 +20,8 @@ You can use `COZY_URL` to connect to another cozy.
 ### `cozy-konnector-dev`
 
 - Provides `COZY_CREDENTIALS` to the locally running connector from `konnector-dev-config.json` (which is created if it does not exist)
+- A token will be generated (`.token.json`) the first time to grant access to `cozy-konnector-dev`, and you may need to delete it and regenerate it again if expired.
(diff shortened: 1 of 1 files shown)


packages/cozy-jobs-cli/README.md
```diff
@@ -20,6 +20,7 @@ You can use `COZY_URL` to connect to another cozy.
 ### `cozy-konnector-dev`
 
 - Provides `COZY_CREDENTIALS` to the locally running connector from `konnector-dev-config.json` (which is created if it does not exist)
+- A token will be generated (`.token.json`) the first time to grant access to `cozy-konnector-dev`, and you may need to delete it and regenerate it again after a while.
```
Crash--, line 23: What is a `while`? Do you mean that the token can be expired? 

While you are at it, I'm wondering if you shouldn't add also that if you change the manifest, you should delete the file too (with a link to this issue) https://github.com/konnectors/libs/issues/701

JF-Cozy: Yes token can be expired.

Oh yes good point for the manifest

Crash--: > it again after a while. 

This sentence is not precise enough to be part of the documentation. You should talk about expired token instead.

Crash-- (approved):

PR #751, opened 2021-02-09
feat: Add new 2FA type (app_code)

LucsT:
Add a new type of 2FA code.


doubleface (approved):

Crash--:
> * @param {string}  options.type - Used by the front to show the right message (email/sms/app)

This doctype string should be updated to reflect this change. 

Also, can you add a bit more context on why this addition? What is the difference between the "app" type?

PR #756, opened 2021-03-01
feat: Export cozy-client models

paultranvan:
This is useful to benefit from the cozy-client models' utilities inside
a konnector.
It is notably required by the tracemob konnector to manipulate time series

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/cozyclient.js
+++ b/packages/cozy-konnector-libs/src/libs/cozyclient.js
@@ -9,6 +9,7 @@
 
 const { Client, MemoryStorage } = require('cozy-client-js')
 const NewCozyClient = require('cozy-client').default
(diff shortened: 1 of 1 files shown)


doubleface:
Isn't it easier to require cozy-client from the connector? I don't see the need to put it in cozy-konnector-libs

paultranvan:
> Isn't it easier to require cozy-client from the connector? I don't see the need to put it in cozy-konnector-libs

As cozy-client is already available from cozy-konnector-libs, it makes more sense to me to expose it from there, and avoid to force the connector to add cozy-client itself, no ?

doubleface:
Ok, I don't see exactly the point. But I do not have better arguments and I don't this this will cause any regression or problem then why not :-)

doubleface (approved):

PR #772, opened 2021-06-02
fix: use es5 build of pdfjs

doubleface:
This fixes some errors with new version of pdfjs

Error: was "The browser/environment lacks native support for critical functionality used by the PDF.js library (e.g. `ReadableStream` and/or `Promise.allSettled`); please use an ES5-compatible build instead."

Source for the fix : https://stackoverflow.com/questions/64189359/reading-pdf-from-url-with-node-js-using-pdf-js

### DIFF
--- a/packages/cozy-konnector-libs/docs/api.md
+++ b/packages/cozy-konnector-libs/docs/api.md
@@ -220,9 +220,9 @@ use this function.
 
 Parameters:
 
-* `documents`: an array of objects corresponding to the data you want to save in the cozy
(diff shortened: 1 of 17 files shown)


Crash--:
> This fixes some errors with new version of pdfjs

Just had an error on something... Doing a git blame on a file and see : "In a few cases, useRef() / ref.current is empty and never re-render. Since this is not the first time you have issues on this, I rewrite the component in Class since it works well." 

But nothing to tell me which cases were not working. Yet this was the most important information to know if my patch had not broken something. 

So I think you should write the 'errors' that this new version fix.

PR #779, opened 2021-07-16
fix: btoa dependency for konnector-dev command

doubleface:
Since last update of cozy-client, a btoa pollyfill is needed for it to work
We use cozy-client for authentication with the konnector-dev command.


LucsT (approved):

Crash--:
Please, let's discuss that here https://github.com/cozy/cozy-client/pull/958 and here https://github.com/cozy/cozy-client/issues/959 

We should work closer to fix this kind of issue

PR #809, opened 2021-10-22
fix: Update README about publishing

LucsT:
Lerna Cli seems to have change a bit, scope option is no more available.
And now we need GH_TOKEN  set up to succeed.

I mention the npx tool too.

What do you think ?


Crash--:
ping @doubleface @LucsT @KillianCourvoisier

doubleface (approved):
Thanks @LucsT

PR #819, opened 2021-11-18
fix: Regression in BaseKonnector unit test

doubleface:
Now COZY_FIELDS needs an "account" attributes to have proper
fields in the BaseKonnector, which is normal after all

This fixes unit test regression introduced in https://github.com/konnectors/libs/pull/816


paultranvan (approved):
Now we need to know why travis failed but successed :confused:

PR #820, opened 2021-11-22
feat: Make peer dependencies optional

doubleface:
Using the peerDependenciesMeta option
https://classic.yarnpkg.com/en/docs/package-json#peerdependenciesmeta-

This will avoid errors like :
not found: Error: Can't resolve 'pdfjs'
when building the connectors which do not use these packages


Crash-- (approved):
Nice thanks!

paultranvan (approved):

PR #821, opened 2021-11-22
fix: Upgrade qualification when metadata already present

LucsT:
Qualification V2 is not updated on existing file if some metadata are already present. (For example carbon_copy)
Will be with this PR.
A test has been added.

A remaining question is, do we want to do the same when all metadata attributes.


LucsT:
We decided to not update all kind of metadata for files stability purpose.

This commit has been tested on a real cozy with a real connector. Merging.

PR #822, opened 2021-11-23
feat: Automatic deploy when merging into master

doubleface:
As other cozy libs do and to allow anybody with proper rights to publish
cozy konnectors libs.

Some github integration will also be added enforce proper pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,23 @@ language: node_js
 matrix:
   fast_finish: true
 node_js:
-- '12'
+- '16'
 cache:
   yarn: true
   directories:
   - node_modules
 script:
-    - lerna bootstrap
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,15 +2,23 @@ language: node_js
 matrix:
   fast_finish: true
 node_js:
-- '12'
+- '16'
```
Crash--, line 5: You should use node 12 and 16. 

By default node 12 every where, but you can run test on node 16 also. 

exemple: https://github.com/cozy/cozy-drive/commit/bb6d22441a3d31a23c5817998c990898b977733e

Crash-- (changes requested):
see comment

PR #829, opened 2022-01-24
feat: Geco and Toni libs integration

doubleface:
- feat: Geco and Toni libs integration
- fix: Buffer deprecations


Crash--:
So we're adding two lib in konnector-libs in order to avoid warnings, right? 

Why did you chose to commit them to the repo instead of forking them?

doubleface:
To avoid the publication step for a small portion of code which did not change since seven years. Do you think it is better to fork here ?

Crash--:
Here we don't know how it's differ from the original libs. This commit https://github.com/konnectors/libs/pull/829/commits/73700f41183157a15e1844610051b6d78fe72e8a explains for Toni but not for Geco. 

You don't need to publish to npm, you can just fork it on github, and then install them form github. 

If there is a force push in https://github.com/rootslab/toni/pull/1 we'll not see the change or else. 
We can't view if our "fork" is up to date with upstream if we add it directly to our codebase 

IMO: there is no good reason to add it to our codebase. ^^

Crash--:
But since this commit fix a lot of noise my comments is not blocking. You can merge it if you don't want / have time to update

doubleface:
I just did the update !

PR #830, opened 2022-01-31
Feat/add error code

Crash--:
We have third party developing software around our konnectors. We need to document all our supported error code. 

from @doubleface advice, I took the list from there https://github.com/cozy/cozy-libs/blob/983f5dc73e50f4a35dab401040c969477e4128ea/packages/cozy-harvest-lib/src/locales/en.json#L124-L207


doubleface (approved):
Thanks @Crash--

PR #837, opened 2022-02-11
feat: Create categorizer options

JulienMirval:
Adds optional parameters to create a categorizer:

- `useGlobalModel` is a boolean which indicates whether to use the global model or not. It prevents depending on the remote asset storing the global model.
- `fetchTransactions` is a function that asynchronously returns an array of transactions, instead of fetching all manually labelled transactions.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/categorization/index.js
+++ b/packages/cozy-konnector-libs/src/libs/categorization/index.js
@@ -13,8 +13,8 @@ const log = logger.namespace('categorization')
 
 /**
  * @typedef CreateCategorizerOptions
- * @property {boolean} useGlobalModel
- * @property {Function} fetchTransactions
+ * @property {boolean} useGlobalModel Whether to use the globally trained model
+ * @property {Function} fetchTransactions A custom training transaction fetcher
  */
 
 /**
@@ -46,7 +46,8 @@ const log = logger.namespace('categorization')
  *   }
  * }
  */
-async function createCategorizer({ useGlobalModel = true, fetchTransactions }) {
+async function createCategorizer(options = {}) {
+  const { useGlobalModel = true, fetchTransactions } = options
   const classifierOptions = { tokenizer }
 
   // We can't initialize the model in parallel using `Promise.all` because with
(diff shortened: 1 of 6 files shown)


packages/cozy-konnector-libs/src/libs/categorization/index.js
```diff
@@ -11,6 +11,12 @@ const logger = require('cozy-logger')
 
 const log = logger.namespace('categorization')
 
+/**
+ * @typedef CreateCategorizerOptions
+ * @property {boolean} useGlobalModel Whether to use the globally trained model
+ * @property {Function} fetchTransactions A custom training transaction fetcher
```
paultranvan, line 17: nit: `fetchTransactions` sounds more boolean than func. Maybe a name `customTransactionFetcher` would make what it does more obvious: it's a fetcher function, and custom (implying it's optional).

packages/cozy-konnector-libs/src/libs/categorization/localModel/classifier.js
```diff
@@ -146,8 +146,16 @@ const reweightModel = classifier => {
 }
 
 const createClassifier = async options => {
-  log('debug', 'Fetching manually categorized transactions')
-  const transactionsWithManualCat = await fetchTransactionsWithManualCat()
+  const { fetchTransactions, ...remainingOptions } = options
+
+  let transactionsWithManualCat = []
```
paultranvan, line 151: The name does not look good anymore: it is not necessarily manual cateogies

paultranvan (approved):
Naming could be improved, but LGTM otherwise :+1:

packages/cozy-konnector-libs/src/libs/categorization/index.js
```diff
@@ -59,26 +59,26 @@ async function createCategorizer(options) {
...
-
   if (useGlobalModel) {
     try {
-      localModel = await createLocalModel({
-        ...classifierOptions,
-        pretrainedClassifier,
-        customTransactionFetcher
-      })
+      globalModel = await createGlobalModel(classifierOptions)
     } catch (e) {
```
paultranvan, line 65: Looks like the first review was not careful enough!

PR #846, opened 2022-02-15
fix: Renovate rule for cozy packages

doubleface:
'after 8pm every weekday' was not recognized by renovate

### DIFF
--- a/packages/renovate-config-cozy-konnector/package.json
+++ b/packages/renovate-config-cozy-konnector/package.json
@@ -19,8 +19,8 @@
           ],
           "timezone": "Europe/Paris",
           "schedule": [
-            "after 8pm every weekday",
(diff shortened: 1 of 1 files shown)


packages/renovate-config-cozy-konnector/package.json
```diff
@@ -19,8 +19,8 @@
           ],
           "timezone": "Europe/Paris",
           "schedule": [
-            "after 8pm every weekday",
-            "before 7am every weekday",
+            "every weekday after 08:00 pm",
+            "every weekday before 07:00 am",
```
trollepierre, line 23: I disagree with your explanation of the solution.

In my understanding, the renovate PRs were opening on time.
But the automerge job was running out of the schedule.

We can try this solution, but I don't think that works

PR #851, opened 2022-03-10
fix: Init cozy-client-js instance from stack env

doubleface:
Also fixes cozy-run-dev command

This way of initializing cozyClient is now not dependent of NODE_ENV but instead of the present of credentials in environment variables.


JulienMirval:
The CozyClientJS was not returned from the `cozyClientJsFromEnv` function

paultranvan (approved):
Thank you! :tada:

PR #852, opened 2022-03-24
fix: Use the right mimeTypes function for content type

doubleface:
If the filename would contain / char, the returned content type was the
filename.
A not expected contentType can cause a lot of problems in our
infrastructure.

Instead, the lookup function expects a complete filename

In parallel, a fix is done in the stack to validate the content type
properly.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/saveFiles.js
+++ b/packages/cozy-konnector-libs/src/libs/saveFiles.js
@@ -389,7 +389,7 @@ async function createFile(entry, options, method, fileId) {
   }
   if (options.contentType) {
     if (options.contentType === true && entry.filename) {
-      createFileOptions.contentType = mimetypes.contentType(entry.filename)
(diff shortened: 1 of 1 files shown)


packages/cozy-konnector-libs/src/libs/saveFiles.js
```diff
@@ -389,7 +389,7 @@ async function createFile(entry, options, method, fileId) {
   }
   if (options.contentType) {
     if (options.contentType === true && entry.filename) {
-      createFileOptions.contentType = mimetypes.contentType(entry.filename)
+      createFileOptions.contentType = mimetypes.contentType(path.extname(entry.filename))
```
sblaisot, line 392: @doubleface returned content-type should probably be checked against `false` and replaced with `application/octet-stream` in that case

```
> mime.lookup(extname("file_with_no_extension"))
false
> mime.lookup("file_with_no_extension")
false
```

PR #859, opened 2022-04-23
Use dynamic import instead of require

### LINKED ISSUE #635: Use es6 imports

kamaradclimber:
This allows to write konnectors as ES module (while still keeping backward compatibility).

Partially fixes #635


Crash--:
ping @doubleface 🙏

PR #870, opened 2022-07-28
feat: Allow connector to read COZY_PAYLOAD

doubleface:
Whether the payload is a JSON string or a file reference

Here is the stack PR for reference : https://github.com/cozy/cozy-stack/pull/3416

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/BaseKonnector.js
+++ b/packages/cozy-konnector-libs/src/libs/BaseKonnector.js
@@ -168,11 +168,16 @@ class BaseKonnector {
    * Read an eventual payload from COZY_PAYLOAD env var, wether it is a JSON string or a reference
    * to a file containing a JSON string
    *
-   * @returns Promise<{Object}> result of JSON.parse from the JSON string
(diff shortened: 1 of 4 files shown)


taratatach (approved):

packages/cozy-konnector-libs/src/libs/BaseKonnector.js
```diff
@@ -163,6 +164,40 @@ class BaseKonnector {
...
+    if (cozyPayload == null) {
+      return null
+    }
+
+    const isFileReference = get(cozyPayload, '[0]') === '@'
+
+    if (isFileReference) {
+      const filePath = cozyPayload.substr(1)
+      try {
+        const fileContent = await fs.readFile(filePath)
```
Crash--, line 185: What does it matter if the file size is 10 or 15MB? Won't it blow up the server's RAM? Do we have an idea of the max size of the BI's payload?

Crash--: @doubleface ? Should we use chunk or something ?

PR #876, opened 2022-08-02
feat(cozy-connector-libs): Remove control character in filename

LucsT:
Remove all control character from filename in saveFiles


doubleface (approved):
Approved when lint is fixed

PR #884, opened 2022-11-02
fix: Do not rely on cwd to read a file payload

doubleface:
Depending on the environment in which the connector is run, we have some problems to read a JSON payload, given as a file by the cozy stack

1) the current working directory with which the connector is run can vary and we cannot depend on it
2) if compiled with webpack with default value like the connector template, the __dirname value is set to "/"

To solve 1) we can use path.resolve(__dirname, filename) to read the payload but this supposes 2) is also solved.

To solve 2), we need to pass the `node: {__dirname: false}` option to webpack. See https://codeburst.io/use-webpack-with-dirname-correctly-4cad3b265a92

The connector template will be updated with the correct webpack configuration

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/BaseKonnector.js
+++ b/packages/cozy-konnector-libs/src/libs/BaseKonnector.js
@@ -10,6 +10,7 @@ const omit = require('lodash/omit')
 const updateOrCreate = require('./updateOrCreate')
 const saveIdentity = require('./saveIdentity')
 const fs = require('fs').promises
+const path = require('path')
 const {
   wrapIfSentrySetUp,
   captureExceptionAndDie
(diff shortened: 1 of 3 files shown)


Crash--:
So even without the `{__dirname: false}` change, using `path.resolve(__dirname, filename)` is enough to read the file's playload given by the stack? If not then the description is a bit misleading. 


If the change in the webpack config is needed, what is your strategy in order to deploy it everywhere?

doubleface:
@Crash-- I updated the description. Is it better ?

doubleface:
> If the change in the webpack config is needed, what is your strategy in order to deploy it everywhere? Shouldn't konnector-libs to handle the webpack config instead of duplicating it on each repo?

A shared webpack configuration is of course the best way but we need to still have the possibility to override webpack configuration. I tried this before and did not manage to make it work properly.

Crash-- (approved):

PR #891, opened 2022-11-29
build: Explicit entry point in webpack

LucsT:
Remove the reference to entry point in package.json, and explicit the index.js used

Work in progress, we need to confirm the good behaviour before merge and auto publishing this package.


LucsT:
This webpack conifg has been tested on a real life case, payfit.
I merge and auto release.

We will still need to activate it on konnectors repos.

Crash--:
Pas sur que d'avoir mis un commit `build` va créer une release automatique ? Lerna n'est pas configuré sur `conventionalCommits` ?

PR #905, opened 2023-01-24
fix: Make option.args optionnal in runInWorkerUntilTrue

doubleface:
To be closer to reality


Crash--:
cc @doubleface

doubleface:
@Crash-- rebased

Crash-- (approved):

PR #914, opened 2023-02-07
BREAKING CHANGE: Adding log level to the log function in ContentScript

LucsT:
This PR report modification on ContentScript.js about the log(level, message) prototype.
We will need to change this.log() call in connectors

### DIFF
--- a/packages/cozy-clisk/src/contentscript/ContentScript.js
+++ b/packages/cozy-clisk/src/contentscript/ContentScript.js
@@ -64,7 +64,8 @@ export default class ContentScript {
     await this.bridge.init({ exposedMethods })
     window.onbeforeunload = () =>
       this.log(
-        'window.beforeunload detected with previous url : ' + document.location
+        'debug',
+        `window.beforeunload detected with previous url : ${document.location}`
       )
 
     this.bridge.emit('workerReady')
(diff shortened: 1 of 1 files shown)


packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -425,10 +425,25 @@ export default class ContentScript {
    * Send log message to the launcher
    *
    * @param {string} message : the log message
```
Crash--, line 427: * @param {"debug"|"info"|"warn"|"error"} level: the log level
🙏

packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -425,10 +425,25 @@ export default class ContentScript {
...
+    }
+    if (!allowedLevels.includes(level)) {
+      level = 'debug'
+    }
+    const now = new Date()
+    this.bridge?.emit('log', {
+      timestamp: now,
+      level,
+      msg: message
+    })
```
Crash--, line 446: Also there is a few call to change: 
https://github.com/konnectors/libs/blob/master/packages/cozy-clisk/src/contentscript/ContentScript.js#L66-L68

for instance

packages/cozy-clisk/src/contentscript/ContentScript.js
LucsT: Fixed

packages/cozy-clisk/src/contentscript/ContentScript.js
LucsT: Done. I leveled this log in debug due to url information.

LucsT:
Made a more 'Conventional commit' message with breaking change in footer

Crash-- (approved):

doubleface (approved):

PR #915, opened 2023-02-07
Feat/ts for clisk

Crash--:
Move some codebase to typescript. The idea is to have a compiler that rejects our errors


packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -54,7 +54,7 @@ export default class ContentScript {
       )
     }
 
-    const exposedMethods = {}
+    const exposedMethods = null
```
Ldoppea, line 57: In the following `for` we have `exposedMethods[method] = ...` so I suppose we'll have an null exception?

packages/cozy-clisk/package.json
```diff
@@ -23,16 +23,18 @@
   "homepage": "https://github.com/konnectors/libs#readme",
   "scripts": {
     "lint": "eslint 'src/**/*.js'",
+    "build:types" : "yarn tsc -p tsconfig.json",
     "prepublishOnly": "yarn run build",
-    "build": "babel src/ -d dist/ --copy-files --verbose --ignore '**/*.spec.js','**/*.spec.jsx'",
+    "build": "yarn build:types && babel --extensions .ts,.js src/ -d dist/ --copy-files --verbose --ignore '**/*.spec.js','**/*.spec.ts'",
```
Ldoppea, line 28: Do we have a check in CI that ensure commited types are aligned with current code?

Crash--: I'll try, but for me, since yarn build:types will return an error code, the pipeline will fail by itself

packages/cozy-clisk/src/bridge/bridgeInterfaces.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+
+
+/**
+ * All messengers are supposed to implement this interface
+ *
+ * @interface
+ */
+export interface MessengerInterface {
```
doubleface, line 36: Now that you converted this MessengerInterface to a real TS interface it is not possible to import it in js code I think. When you look at the /dist/bridge/bridgeInterfaces.js it builds as an empty file.

Should we migrate flagship app files using theses interfaces to TS ?

LucsT: I think we should because we need some changes present here for the launcher to work.
Ex : changes in addData 
```
- import CozyClient from 'cozy-client/types/CozyClient'
...
- * @param {CozyClient} options.client : CozyClient instance
+ * @param {import('cozy-client/types/CozyClient').default} options.client CozyClient instance
```

packages/cozy-clisk/src/launcher/addData.js
```diff
@@ -12,7 +11,7 @@ const log = Minilog('addData')
  * @param {Array} entries : an array of objects corresponding to the data you want to save in the cozy
  * @param {string} doctype : the doctype where you want to save data (ex: 'io.cozy.bills')
  * @param {object} options : options object
- * @param {CozyClient} options.client : CozyClient instance
+ * @param {import('cozy-client/types/CozyClient').default} options.client CozyClient instance
```
LucsT, line 14: About this change in addData, and other files.
I have build and linked correctly to flagship app the lib.
I confirm it's working now with the konnector template. And i can delete addData, hydrateAndFilter, saveFiles, saveBills, saveIdentity

packages/cozy-clisk/src/bridge/bridgeInterfaces.ts
Crash--: I'll split the PR in order to quickly release a fix.

packages/cozy-clisk/src/contentscript/ContentScript.js
Ldoppea: You should be able to do this
```js
/** @type {import('post-me').MethodsType} */
const exposedMethods = {}
```

packages/cozy-clisk/src/bridge/bridgeInterfaces.ts
Crash--: > Should we migrate flagship app files using theses interfaces to TS ?

@doubleface To me, this is not needed. We can do it, but we can also do differently: 
- You can remove the `extends` keyword for the class' on the flagship side
- Return the correct type (jsdoc) for these classes 

In all cases, we need to rewrite the classes, since `clisk` should not publish any code related to the interface.

PR #916, opened 2023-02-10
Fix/js doc

Crash--:
Subset of #915 in order to get JSDoc fix to be able to use it on the flagship side.


LucsT:
Builded and tested with template

LucsT (approved):

PR #917, opened 2023-02-14
feat(cozy-clisk): Add a single entry point

LucsT:
This PR make a single entry point for the cozy-clisk lib to successfully import bridge interfaces in flagship app.

Todo : 
When use in connectors this new lib version will break this import : https://github.com/konnectors/alan/blob/beeabab21782b96fc5d7018dacfc2b74d2f07343/src/index.js#L2


LucsT:
Will be use in this PR https://github.com/cozy/cozy-flagship-app/pull/599

Crash--:
Since this commit is a breaking change, you should add this information in the commit message with the way to handle it. https://github.com/cozy/cozy-guidelines/#breaking-change

Crash-- (approved):

PR #919, opened 2023-02-23
fix: convert date string to date object in saveBills

doubleface:
Since the initial date object has been converted to string with post-me
serialization

We still check if the date string is convertible to date.

Also added some corresponding unit tests to saveBills


Crash-- (approved):

LucsT (approved):
Tested with Alan

PR #920, opened 2023-02-28
Feat/download file in worker

LucsT:
This PR implement the fonction that will call by launcher to download file in worker.

We remove the download during saveFiles and the filtering logic.
Filtering logic launcher side, will be implemented in the next step.

Will be used in app by https://github.com/cozy/cozy-flagship-app/pull/633

### DIFF
--- a/packages/cozy-clisk/src/contentscript/ContentScript.js
+++ b/packages/cozy-clisk/src/contentscript/ContentScript.js
@@ -27,7 +27,7 @@ export default class ContentScript {
    *
    * @param {object} options : options object
    * @param {Array<string>} [options.additionalExposedMethodsNames] : list of additional method of the
-   * content script to expose expose. To make it callable via the worker
+   * content script to expose. To make it callable via the worker.
    */
   async init(options = {}) {
     this.bridge = new LauncherBridge({ localWindow: window })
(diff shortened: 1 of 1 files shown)


Crash--:
Please give us some context on this PR. Why did you do this change?

LucsT:
I have do this change in this card context  : https://trello.com/c/ZSEywYCe/229-%F0%9F%93%A5-perf-t%C3%A9l%C3%A9charger-les-fichiers-au-fur-et-%C3%A0-mesure

The final goal is to be able to let the launcher decided which file we need to download (Website  -> App) to reduce the number of file to download each time we execute the connector.

The first step is to be able to download the file in worker. 
This is the feature added by this PR and https://github.com/cozy/cozy-flagship-app/pull/633
Exactly, this PR bring to the contentScript (so the konnector) the ability to download with this function.


The second step will based on this and will add the filtering based on filename and metadata in the saveFiles of the flagship app.

Do you need more context or details ?

LucsT:
Not a draft anymore, I 'publish' the PR.

doubleface:
@LucsT Please add this context in the main commit and it will be ok for me

LucsT:
Enhance commit body

doubleface (approved):

PR #923, opened 2023-03-13
Feat/get classifier options

JulienMirval:
`getClassifierOptions` only needs the number of classes to get the options, we were providing the full array of transactions. This is a more general approach that will fit other use cases.

We took the liberty of NOT marking the breaking change because it affects a previously unexported method that is not used anywhere but in the DisseCozy POC (c.f. #921)

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/categorization/localModel/classifier.js
+++ b/packages/cozy-konnector-libs/src/libs/categorization/localModel/classifier.js
@@ -36,8 +36,13 @@ const TOKENS_TO_REWEIGHT = [
   'tag_pos tag_activity_income'
 ]
 
-const getClassifierOptions = uniqueCategories => {
-  const nbUniqueCategories = uniqueCategories.length
+/**
+ * Get the classifier options, mainly to get the alpha parameter
+ *
+ * @param {number} nbUniqueCategories - Number of unique categories
+ * @returns {object} the classifier options
+ */
+const getClassifierOptions = nbUniqueCategories => {
   log(
     'debug',
     'Number of unique categories in transactions with manual categories: ' +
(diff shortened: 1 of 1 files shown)


packages/cozy-konnector-libs/src/libs/categorization/localModel/classifier.js
```diff
@@ -36,8 +36,7 @@ const TOKENS_TO_REWEIGHT = [
   'tag_pos tag_activity_income'
 ]
 
-const getClassifierOptions = transactionsWithManualCat => {
-  const uniqueCategories = getUniqueCategories(transactionsWithManualCat)
+const getClassifierOptions = uniqueCategories => {
```
paultranvan, line 39: By reading the code, I think you can just pass the length of the categories rather than the actual array.
Also, could you add some jsdoc? Like: 
```
/**
 * Get the classifier options, mainly to get the alpha parameter 
 *
 * @param {number} nUniqueCategories - Number of unique categories
 * @returns {object} the classifier options
 */
 ```

File: packages/cozy-konnector-libs/src/libs/categorization/localModel/classifier.js [hunk 2]
```diff
@@ -166,7 +170,8 @@ const createClassifier = async options => {
 
   log('debug', 'Instanciating a new classifier')
 
-  const classifierOptions = getClassifierOptions(transactions)
+  const nbUniqueCategories = getUniqueCategories(transactions)
+  const classifierOptions = getClassifierOptions(nbUniqueCategories.length)
```
paultranvan, line 174: You might want to have more defensive code, e.g. `getClassifierOptions(a?.length || 0)`

PR #927, opened 2023-03-16
fix: Babel transpilation

Crash--:
It seems that when we publish the mono-repo we're running our Babel compilation process from within the subpackages.

By doing so, babel is not aware of the root configuration files. So the babel config doesn't apply.

To fix that we need to pass `--root-mode upward` which will make Babel search from the working directory upward looking for the babel.config.json

See https://babeljs.io/docs/config-files#root-babelconfigjson-file for babel documentation

This commit fix the issue where we had an optional chaining in our dist/ folder.


paultranvan (approved):
Thanks a lot!

doubleface (approved):
Good catch thanks !

PR #929, opened 2023-03-16
fix: Remove cozy-client-js dependency in cozy-clisk

doubleface:
Since it is no longer used


LucsT (approved):
Approved, but maybe you should run yarn to update yarn.lock before merging.

doubleface:
@LucsT I did it, but since it is a monorepo, cozy-client-js is still needed by cozy-konnector-libs and it does not change yarn.lock.

But the published cozy-clisk won't have cozy-client-js

PR #930, opened 2023-03-17
feat[cozy-clisk]:Implementing selective file dowload and fixing metadata dedupli…

LucsT:
This PR bring a new exported function in cozy-clisk for flagship app.
This function is needed by the launcher to evaluate early if the file can be found on the cozy.
This code will be use by this PR on flagship app [TO BE ADDED]

-> Separation and  adaptation of function getFileIfExist in the lib
-> Evaluation of some parameters sooner in the process (folderpath, sourceAccountIdentifier, ...)
-> Adding some units tests that cover the code pretty well.
-> Fixing deduplication on metadata that was not operational before due to received data structure

Need to be done before merge
-> Adapting timing wraper to this new lib function.

Should be done soon
-> Moving 'shouldreplace' sooner in the process
-> We lost the 'free' metadata update when file is already present, as we now do not update each time the file. Need to be adapt. (ex: add qualification on an existing already download file, ...)


LucsT:
Rebase and cleaned.

I still have need to add 3 @ts-ignore 
I don't know of to easily fix it for now, and they are linked to externals component of the lib.
What do you think about it?

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
+import Minilog from '@cozy/minilog'
+import { Q } from 'cozy-client'
+import get from 'lodash/get'
+
+const log = Minilog('getFileIfExists')
+
+module.exports = getFileIfExists
+
+async function getFileIfExists({ client, entry, options, folderPath, slug }) {
```
Crash--, line 9: jsdoc please with `//ts-check`

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+  const fileIdAttributes = options.fileIdAttributes
+  const sourceAccountIdentifier = options.sourceAccountIdentifier
```
Crash--, line 11: nit: const { fileIdAttributes, sourceAccountIdentifier } = options

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+      // path from a previous version of the connector
+      log.debug('Rolling back on detection by filename')
+      return getFileFromPath(client, entry, folderPath)
+    } else {
+      return file
+    }
+  } else {
+    log.debug('Rolling back on detection by filename')
+    return getFileFromPath(client, entry, folderPath)
+  }
```
Crash--, line 35: We prefer the early return, it's a lot easier to read: 
```
if(!isReadyForFileMedata){
return getFileFromPath
} else {
... 
...
} 
```

packages/cozy-clisk/src/launcher/getFileIfExists.js
Crash--: If we make a request to the cozy, then the method should be called fetchSomething (https://github.com/cozy/cozy-guidelines/#naming-of-functions)

Get should only be used to get something in memory

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+}
+
+async function getFileFromMetaData(
```
Crash--, line 38: same here for naming convention. Please follow https://github.com/cozy/cozy-guidelines/#naming-of-functions

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+
+module.exports = getFileIfExists
+
+async function getFileIfExists({ client, entry, options, folderPath, slug }) {
+  const fileIdAttributes = options.fileIdAttributes
+  const sourceAccountIdentifier = options.sourceAccountIdentifier
+
+  const isReadyForFileMetadata =
+    fileIdAttributes && slug && sourceAccountIdentifier
+  if (isReadyForFileMetadata) {
```
Crash--, line 15: to me `isReadyForFileMetadata` means that this method will be called several times in a short period of time. But I don't think this is the meaning here. It's more `hasEnoughtMetadata` or something no?

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+          createdByApp: slug
+        }
+      })
+      .indexFields([
+        'metadata.fileIdAttributes',
+        'trashed',
+        'cozyMetadata.sourceAccountIdentifier',
+        'cozyMetadata.createdByApp'
+      ])
+  )
```
Crash--, line 66: 1/ We should really avoid the queryAll() method. This method can return a lot of documents and can be pretty long to execute. Can't we restrict the returned data by querying only the last 12 months or else? 

2/ I don't know if you need all the fields from this `io.cozy.files` but you can reduce the size of the returned data by selecting them by adding : `.select(['name', 'cozyMetadata.createdAt'])`

3/ You queryAll() but after that, you only return the first result. You should use something like this instead: 
```
client.query(Q('io.cozy.files').where...limitBy(1))
```

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+async function getFileFromMetaData(
+  client,
+  entry,
+  fileIdAttributes,
+  sourceAccountIdentifier,
+  slug
+) {
+  log.debug(
+    `Checking existence of ${calculateFileKey(entry, fileIdAttributes)}`
+  )
```
Crash--, line 47:
```
const fileKey = calculateFileKey(entry, fileIdAttributes)
```

and use fileKey later instead of calculating it 3 times

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+        cozyMetadata: {
+          sourceAccountIdentifier,
+          createdByApp: slug
+        }
+      })
+      .indexFields([
+        'metadata.fileIdAttributes',
+        'trashed',
+        'cozyMetadata.sourceAccountIdentifier',
+        'cozyMetadata.createdByApp'
```
Crash--, line 64: Since the order is important (https://docs.cozy.io/en/tutorials/data/advanced/#indexes-performances-and-design) I'll changed to: 

```
[
        'metadata.fileIdAttributes',
  'cozyMetadata.sourceAccountIdentifier',
        'cozyMetadata.createdByApp'
        'trashed'
      
]
```

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+      )
+    }
+    return files[0]
+  } else {
+    log.debug('File not found')
+    return false
+  }
+}
+
+async function getFileFromPath(client, entry, folderPath) {
```
Crash--, line 83: jsdoc 🙏

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+  try {
+    log.debug(`Checking existence of ${getFilePath({ entry, folderPath })}`)
```
Crash--, line 85: store the result of getFilePath and use it later 🙏

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+        `Found ${files.length} files corresponding to ${calculateFileKey(
+          entry,
+          fileIdAttributes
+        )}`
+      )
+    }
+    return files[0]
+  } else {
+    log.debug('File not found')
+    return false
```
Crash--, line 79: should return null. Not boolean

packages/cozy-clisk/src/launcher/getFileIfExists.js
```diff
@@ -0,0 +1,128 @@
...
+async function getFileFromPath(client, entry, folderPath) {
+  try {
+    log.debug(`Checking existence of ${getFilePath({ entry, folderPath })}`)
+    const result = await client
+      .collection('io.cozy.files')
+      .statByPath(getFilePath({ entry, folderPath }))
+    return result.data
+  } catch (err) {
+    log.debug(err.message)
+    return false
```
Crash--, line 92: same should return null

packages/cozy-clisk/src/launcher/getFileIfExists.js
paultranvan: I'd also suggest to use a partial index for the trashed attribute: 
```
    Q('io.cozy.files')
      .where({
        metadata: {
          fileIdAttributes: calculateFileKey(entry, fileIdAttributes)
        },
        cozyMetadata: {
          sourceAccountIdentifier,
          createdByApp: slug
        }
      })
      .indexFields([
        'metadata.fileIdAttributes',
        'cozyMetadata.sourceAccountIdentifier',
        'cozyMetadata.createdByApp'
      ])
      .partialIndex({
        trashed: false
      })

```

Crash--: Why that? I though the same at first (putting it in a partial index) but then I checked `trashed` attribute and it seems that is always there. So I don't understand why to put it in a partial index. :'(

paultranvan: Since it's always there, it will always be indexed, but we only need the `trashed: false` docs. So the partial index will allow a lighter index with only non-trashed files

Crash--:
Y a un truc qui me dérange avec ça, c'est qu'on fait les requêtes fichier par fichier. Et si on en a beaucoup, ça va vraiment pas être terrible.

Crash--:
Et aussi, on est censé avoir un context sur les clisk des fichiers déjà récupérés etc, on devrait le passer à cet endroit là, non ?

doubleface:
> @Crash-- Et aussi, on est censé avoir un context sur les clisk des fichiers déjà récupérés etc, on devrait le passer à cet endroit là, non ?

On avait enlevé la récupération du contexte lorsque les fichier n'avaient pas le bon createdByApp. Et on ne l'a pas remis depuis. Je pense qu'on peut faire une carte pour ça

doubleface:
> @Crash-- Y a un truc qui me dérange avec ça, c'est qu'on fait les requêtes fichier par fichier. Et si on en a beaucoup, ça va vraiment pas être terrible.

Avec un index ça n'est pas sensé être suffisamment performant ? Parce que sinon, on va de toutes façons devoir se reconstruire un inde x en mémoire pour savoir si un fichier existe déjà.

En même temps, c'est vrai qu'ici on est sur un téléphone et le temps de réponse du réseau risque d'être plus élevé. Je ne sais pas vraiment dire à quel point.

Mais je suppose que c'est une bonne chose qu'on soit plus facilement capables de faire des mesures maintenant

packages/cozy-clisk/src/launcher/getFileIfExists.js
doubleface: Yes, the queryAll is here only to know if more than one document satisfies the request, to log a waring if more than on file does it.

My mistake in the first place.

The get the same feature, we should then use client.query with limitBy(2)

Crash--:
@doubleface @LucsT what is the status of this PR?

PR #931, opened 2023-03-20
feat: Migrate statByPath in cozy-konnector-libs

LucsT:


Crash--:
The best approach would certainly be to create this mkdirp method directly in cozy-client/models/file.js.

doubleface:
Actually, we should just deprecate mkdirp and use [createDirectoryByPath](https://github.com/cozy/cozy-client/blob/e4cf003cb548714d8b2c999578ee87cb4e82ce97/packages/cozy-stack-client/src/FileCollection.js#L896) instead, which does the job

Crash--:
Oh nice, we already have this one! 👏 

You can also branch mkdirp to createDirectoryByPath. Like that, no need to deprecated and change konnector code.

Crash--:
@LucsT what is the status of this PR?

PR #936, opened 2023-03-27
feat: Enhance timeout error messages

doubleface:
Now when a timeout error is thrown, it will have more information about
the caller in the error message.

### DIFF
--- a/packages/cozy-clisk/jest.config.js
+++ b/packages/cozy-clisk/jest.config.js
@@ -1,4 +1,5 @@
 module.exports = {
   testEnvironment: 'jsdom',
-  testPathIgnorePatterns: ['/node_modules/', '/dist/']
+  testPathIgnorePatterns: ['/node_modules/', '/dist/'],
+  transformIgnorePatterns: ['/node_modules/(?!(p-timeout|p-wait-for))']
 }


packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -235,7 +237,12 @@ export default class ContentScript {
...
-      timeout: DEFAULT_WAIT_FOR_ELEMENT_TIMEOUT
+      timeout: {
+        milliseconds: DEFAULT_WAIT_FOR_ELEMENT_TIMEOUT,
+        message: new TimeoutError(
+          `waitForElementNoReload ${selector} timed out after ${DEFAULT_WAIT_FOR_ELEMENT_TIMEOUT}ms`
+        )
+      }
     })
     return true
   }
```
Crash--, line 248: Nice, but how this error can't be converted to a job error and stop the execution of the konnector?

doubleface: @Crash-- It would be nice to merge this PR to have better timeout error messages, even if it does not fix any problem. Can we merge it ?

Crash-- (approved):

PR #937, opened 2023-03-27
feat: Allow waitForElementInWorker to set its own timeout

doubleface:
The default timeout for `waitForElementInWorker` was previously 30s. Sometimes, when the website is slow, en element may take more than 30s to display.

It is now possible to specify the timeout of this command, to have a longer or shorter timeout.

Now waitForElementInWorker can be called like this : 

```javascript
this.waitForElementInWorker(selector, {
  timeout: 100000 // for 100s
})
```


KillianCourvoisier (approved):

Crash--:
What was the previous default? Why is it needed? Can you add context please?

doubleface:
@Crash-- PR description now updated with context

PR #940, opened 2023-04-21
 feat: Allow to watch and update a konnector on local cozy stack 

doubleface:
With this, if you update cozy-konnector-build on your konnector (clisk
or node), you can have the konnector on you local stack updated in
realtime for each modification of your konnector or its dependencies.

The command :

```
cozy-stack konnectors update <slug from current manifest>
```

is run after each source code modification.

The targetted cozy-stack instance is chosen by the cozy-stack command
with the $COZY_DOMAIN environment variable and targets
http://cozy.localhost:8080/ by default.

To make it easier to use, add

```javascript
{
  "scripts": {
    "watchDeploy": "UPDATE_ON_LOCAL_COZY_STACK=true yarn watch"
  }
}
```

to your konnector's package.json file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ packages/*/node_modules/*
 yarn-error.log
 packages/*/dist/*
 .changelog/*
+**/.vscode/*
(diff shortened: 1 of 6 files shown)


doubleface:
Closed since now the cozy stack does not need an update when the source is in the form `file://`

PR #941, opened 2023-04-21
feat: Add evaluateInWorker method

doubleface:
This method allow to evaluate a function in worker context, without
having to declare a method in ContentScript before.

Example:

```javascript
const url = await this.evaluateInWorker(() => document.location.href)
```

This may of course be a little less optimized than declaring a method
before, but it is easier to follow I think.

This method can take :
 - normal functions
 - arrow function

Function returning a Promise will be awaited

```javascript
const result = await this.evaluateInWorker(() => {
  return new Promise(resolve =>
    window.setTimeout(() => resolve(window.document.location.href), 1000)
  )
})
```

Async function are not handled yet because of a babel transform problem
at the moment.


KillianCourvoisier (approved):
Could be useful indeed 👍

PR #943, opened 2023-05-22
fix: Deduplicate identities on createdByApp

doubleface:
cozyMetadata.createdByApp is not present anymore in the document given
to saveIdentity.

We have to use the slug saved in client.appMetadata instead to
deduplicated identities.


Crash-- (changes requested):
Let's use cozy-client to fix this issue 🙏

packages/cozy-clisk/src/launcher/saveIdentity.js
```diff
@@ -54,6 +54,15 @@ export default async (contactOrIdentity, accountIdentifier, options = {}) => {
...
 
+  // force cozyMetadata.createdByApp into the document to allow updateOrCreate to deduplicate
+  // identities with correct data. createdByApp will be forced by CozyClient anyway
+  const client = options?.client
+  if (client.appMetadata?.slug) {
+    identity.cozyMetadata = {
+      createdByApp: client.appMetadata?.slug
+    }
+  }
+
```
Crash--, line 65: cozy-client will add the createdByApp automatically. We should not add it manually!

Crash--: By the way, server side konnector has done the half way to migrate to cozyclient https://github.com/konnectors/libs/blob/master/packages/cozy-konnector-libs/src/libs/saveIdentity.js#L82-L116

doubleface: I know, but cozyMetadata.createdByApp will be added by CozyClient once the document is added to the database. Here, we want to compare the existing documents in database with the document we are trying to add, and the document we are trying to add does not have cozyMetadata.createdByApp yet.

But it was a naive quick solution. I will propose another one which does not use updateOrCreate since this is not the way to go.

doubleface: @Crash-- PR is updated and ready and now I use cozy-client directly. This is better this way

PR #947, opened 2023-06-09
feat: Add a function to check if element is present in the worker

KillianCourvoisier:
In order to avoid multiple checking functions for elements in the worker, we add a generic function to be use in every konnector

### DIFF
--- a/packages/cozy-clisk/src/contentscript/ContentScript.js
+++ b/packages/cozy-clisk/src/contentscript/ContentScript.js
@@ -230,6 +230,17 @@ export default class ContentScript {
   }
 
   /**
+   * Check if dom element is present on the page.
+   *
+   * @param {string} selector - css selector we are checking for
+   * @returns {Promise<boolean>}  - Returns true or false
+   */
+  async isElementInWorker(selector) {
+    this.onlyIn(PILOT_TYPE, 'isElementInWorker')
+    return await this.runInWorker('checkForElement', selector)
+  }
+
+  /**
    * Wait for a dom element to be present on the page. This won't resolve if the page reloads
    *
    * @param {string} selector - css selector we are waiting for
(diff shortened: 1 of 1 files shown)


packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -229,6 +229,17 @@ export default class ContentScript {
...
 
+  /**
+   * Check if dom element is present on the page.
+   *
+   * @param {string} selector - css selector we are checking for
+   * @returns {boolean} - Returns true or false
+   */
+  isElementInWorker(selector) {
+    this.onlyIn(PILOT_TYPE, 'isElementInWorker')
+    return Boolean(this.checkForElement(selector))
```
doubleface, line 240: A runInWorker is missing for this to work properly

doubleface (changes requested):

packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -229,6 +229,17 @@ export default class ContentScript {
...
 
+  /**
+   * Check if dom element is present on the page.
+   *
+   * @param {string} selector - css selector we are checking for
+   * @returns {Promise<boolean>}  - Returns true or false
+   */
+  async isElementInWorker(selector) {
+    this.onlyIn(PILOT_TYPE, 'isElementInWorker')
+    return Boolean(await this.runInWorker('checkForElement', selector))
```
doubleface, line 240: No need for a boolean conversion here since checkForElement already does that.

packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -249,6 +260,18 @@ export default class ContentScript {
     return true
   }
 
+  /**
+   * Check if a dom element is present on the page.
+   *
+   * @param {string} selector - css selector we are waiting for
```
doubleface, line 266: nit: Please fix this comment since we do not wait for this element here

PR #949, opened 2023-06-14
feat: Download files progressively only when needed 

doubleface:
Now the launcher will only have to call saveFiles directly for all the
given files,  which will avoid to download all the files before calling
saveFiles. (PR in flagship app to follow)

The downloadAndFormatFile option callback has been added to allow
saveFiles to download the corresponding file in the worker when needed

For optimization, an index of already downloaded files is created on
each call to saveFiles. Then, when saveFiles is called with multiple
entries, the need to download a file will be detected quickly.

Also clean some code and added jsdoc as much as possible now.

### DIFF
--- a/packages/cozy-clisk/src/launcher/saveFiles.js
+++ b/packages/cozy-clisk/src/launcher/saveFiles.js
@@ -104,14 +104,6 @@ const saveFiles = async (client, entries, folderPath, options) => {
     existingFilesIndex: options.existingFilesIndex
   }
 
-  if (options.validateFileContent) {
-    if (options.validateFileContent === true) {
-      saveOptions.validateFileContent = defaultValidateFileContent
-    } else if (typeof options.validateFileContent === 'function') {
-      saveOptions.validateFileContent = options.validateFileContent
-    }
-  }
-
   noMetadataDeduplicationErrors(saveOptions)
 
   let savedFiles = 0
@@ -315,14 +307,6 @@ async function createFile(client, entry, options, method, file) {
       await removeFile(client, fileDocument)
       throw new Error('BAD_DOWNLOADED_FILE')
     }
-
-    if (
-      options.validateFileContent &&
-      !(await options.validateFileContent(fileDocument))
-    ) {
-      await removeFile(client, fileDocument)
-      throw new Error('BAD_DOWNLOADED_FILE')
-    }
   }
 
   return fileDocument
@@ -468,14 +452,6 @@ function calculateFileKey(entry, fileIdAttributes) {
 
 function defaultValidateFile(fileDocument) {
   return checkFileSize(fileDocument)
-}
-
-async function defaultValidateFileContent(fileDocument) {
-  if (!defaultValidateFile(fileDocument)) {
-    log.warn('Wrong file type from content')
-    return false
-  }
-  return true
 }
 
 function sanitizeEntry(entry) {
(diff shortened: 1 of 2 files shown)


packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -154,29 +164,17 @@ const saveEntry = async function (client, entry, options) {
...
-    options,
-    'sourceAccountOptions.sourceAccountIdentifier'
-  )
-  if (!sourceAccountIdentifier) {
-    log.warn(
-      'No sourceAccountIdentifier is defined in options, file deduplication will be based on file path'
-    )
+    throw new Error('Savefiles : No slug is defined for the current connector')
   }
 }
```
Crash--, line 250: should be tagged as a BREAKING CHANGE even if you think that all clisk will not be impacted by this change.

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -504,6 +464,15 @@ function attachFileToEntry(entry, fileDocument) {
...
 
+/**
+ * Get the full path of a given file, from it's folder path and file or entry
+ *
+ * @param {Object} arg - arg option object
+ * @param {import('cozy-client/types/types').FileDocument} [arg.file] - io.cozy.files object
+ * @param {saveFilesEntry} [arg.entry] - saveFiles entry
+ * @param {saveFileOptions} arg.options - io.cozy.files object
+ * @returns {string | undefined} - file full path
```
Crash--, line 474: this is not part of this PR, but its weird to return undefined from this kind of method.

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -1,23 +1,89 @@
+// @ts-check
 import Minilog from '@cozy/minilog'
 import get from 'lodash/get'
+import set from 'lodash/set'
```
Crash--, line 4: Can't we avoid the use of lodash set? 

Remember: We should go away from lodash. It'll reduce the size of the content script ;)

Crash-- (approved):
Once commits are squashed / removed, let's merge :)

packages/cozy-clisk/src/launcher/saveFiles.js
doubleface: It is allways possible but a lot less elegant and readable I think. But let's do this.

packages/cozy-clisk/src/launcher/saveFiles.js
doubleface: Done

packages/cozy-clisk/src/launcher/saveFiles.js
doubleface: Done

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -104,14 +104,6 @@ const saveFiles = async (client, entries, folderPath, options) => {
     existingFilesIndex: options.existingFilesIndex
   }
```
Crash--, line 106: this commit should be tagged as a breaking change

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -7,17 +7,27 @@ import retry from 'bluebird-retry'
 const log = Minilog('saveFiles')
```
Crash--, line 8: This is not the right semantic see https://github.com/cozy/cozy-guidelines/#breaking-change 🙏

doubleface: Now it should be better

Crash--: Nop. No space before the `:`

doubleface: This time is the right one

PR #954, opened 2023-06-29
feat: Add a onWorkerReady callback to the content script

doubleface:
This method of the content script will be called each time the worker
page is reloaded and when the worker bridge is ready to take commands.

This will be useful to subscribe to page events and redirect it to pilot
methods.

Ex :

```javascript
  onWorkerReady() {
    const button = document.querySelector('input[type=submit]')
    if (button) {
      button.addEventListener('click', () =>
        this.bridge.emit('workerEvent', 'loginSubmit')
      )
    }
  }
```

The `workerEvent` event will be sent directly to the pilot and here the
pilot will be able to call `this.blockWorkerInteractions` for example.


KillianCourvoisier (approved):
This will be useful indeed, thanks 🙏

PR #960, opened 2023-07-28
feat: Add waitForNotAuthenticated shortcut method to ContentScript

doubleface:
This method is needed in almost every ensureNotAuthenticated and is
better shared here

Added timeout and interval options since it can be different depending
on the tiers site. And for conssistency, I added the same option to
waitForAuthenticated.

### DIFF
--- a/packages/cozy-clisk/src/contentscript/ContentScript.js
+++ b/packages/cozy-clisk/src/contentscript/ContentScript.js
@@ -170,17 +170,48 @@ export default class ContentScript {
    * This method is made to run in the worker and will resolve as true when
    * the user is authenticated
    *
+   * @param {object} options        - options object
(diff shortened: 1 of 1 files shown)


KillianCourvoisier (approved):

packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -170,17 +170,48 @@ export default class ContentScript {
...
+          `waitForAuthenticated timed out after ${timeout}ms`
+        )
+      }
+    })
+    return true
+  }
+
+  /**
+   * This method is made to run in the worker and will resolve as true when
+   * the user is authenticated
```
KillianCourvoisier, line 197: is not authenticated* right ?

doubleface: right

PR #968, opened 2023-10-04
fix: Add full anticaptcha response possibility

LucsT:
Add possibility to have user agent in full anticaptcha response with solveCaptchaWithFullResponse


packages/cozy-konnector-libs/src/libs/solveCaptchaWithFullResponse.js
```diff
@@ -0,0 +1,179 @@
...
+ * Parameters:
+ *
+ * - `params` is an array of objects with any attributes with some mandatory attributes :
+ *   + `type` (String): (default recaptcha) type of captcha to solve. can be "recaptcha" or "image" at the moment
+ *   + `timeout` (Number): (default 3 minutes after now) time when the solver should stop trying to
+ *   solve the captcha
+ *   + `websiteKey` (String): the key you can find on the targeted website (for recaptcha)
+ *   + `websiteURL` (String): The URL of the page showing the captcha (for recaptcha)
+ *   + `body` (String): The base64 encoded image (for image captcha)
+ * Returns: Promise with the solved captcha response as a string
```
Crash--, line 34: nice doc, you should definitly write that within the JSDoc format! It'll be useful for our IDE.

packages/cozy-konnector-libs/src/libs/solveCaptchaWithFullResponse.js
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * @example
+ *
+ * ```javascript
+ * const { solveCaptchaWithFullResponse } = require('cozy-konnector-libs')
+ *
+ * const solvedKey = await solveCaptchaWithFUllResponse({
```
Crash--, line 41:
```suggestion
 * const solvedKey = await solveCaptchaWithFullResponse({
```

packages/cozy-konnector-libs/src/index.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
   normalizeFilename: require('./libs/normalizeFilename'),
   utils: require('./libs/utils'),
   solveCaptcha: require('./libs/solveCaptcha'),
```
Crash--, line 39: Do we still need to expose this `solveCaptcha`? Shoudn't we use only `solveCaptchaWithFullResponse` ?

Also, I don't understand the `withFullResponse` since we still return `solveWithAntiCaptcha`. And this `fullResponse` mean?

Crash--: In that case, why not editing the signature of the `solveCaptcha` method to something like: 

```
solveCaptcha(params, fullResponse = false)
```

By default, solveCaptcha will continue to return the same thing as before, but you can get the fullResponse by setting this arg to true. Like that : only one method to expose. 

If the second arg is not a good option, you can also pass the information to the first param object. 

I prefer to have as small an API surface as possible instead of creating new methods. It'll be easier for a dev to know what to do. (and for us to maintain)

LucsT: We thought about adding an argument.
That's was the first choice but after talking with @doubleface with think it's cleaner to have 2 method with different kind of return.

We need to talk about it.

doubleface: I found it better to enforce the right api, which could return an object with possibly  multiple keys instead of a string.

Using fullResponse parameter works but enforces an api which can return different types of responses depending on parameters, which is less readable to my opinion.

Crash--: This is where inferred types are great. Depending of what args you pass, you get the right return type. Our API is way too big today, we expose too more functions and it's hard to know.

Having a method that return <string, string[]> is not bad, is it?

doubleface: It looked to me as a bad practice to return different types of data depending on the parameters.
But let's do this. And it will be simpler to implement.

PR #973, opened 2023-11-10
feat: Now create destination folders before saving files

doubleface:
And folders are created only for files which will be downloaded (this avoids creating empty folders when files already exist elsewhere).

This avoids many requests to the stack for each imported files and saves a lot of time

If the destination folder was removed, throw a specific error to let the
flagship app handle the situation (recreate the destination folder,
update the trigger with folder_to_save, etc)

I finally did not need any folder map, since cozy-client has it's own with [ensureDirectoryExists](https://github.com/cozy/cozy-client/blob/81e6523fab6fc927e6526d880910608124168059/packages/cozy-stack-client/src/FileCollection.js#L865-L871)

### DIFF
--- a/packages/cozy-clisk/src/launcher/saveFiles.js
+++ b/packages/cozy-clisk/src/launcher/saveFiles.js
@@ -25,7 +25,6 @@ import { dataUriToArrayBuffer } from '../libs/utils'
  * @property {Array<string>} fileIdAttributes - List of entry attributes considered as unique deduplication key
  * @property {Function} log - Logging function coming from the Launcher
  * @property {string} [subPath] - subPath of the destination folder path where to put the downloaded file
- * @property {Function} [postProcess] - callback called after the file is download to further modify it
  * @property {string} [contentType] - will force the contentType of the file if any
  * @property {Function} [shouldReplaceFile] - Function which will define if the file should be replaced or not
  * @property {Function} [validateFile] - this function will check if the downloaded file is correct (by default, error if file size is 0)
@@ -41,7 +40,6 @@ import { dataUriToArrayBuffer } from '../libs/utils'
  * @property {Array<string>} fileIdAttributes - List of entry attributes considered as unique deduplication key
  * @property {Function} log - Logging function coming from the Launcher
  * @property {string} [subPath] - subPath of the destination folder path where to put the downloaded file
- * @property {Function} [postProcess] - callback called after the file is download to further modify it
  * @property {string} [contentType] - will force the contentType of the file if any
  * @property {Function} [shouldReplaceFile] - Function which will define if the file should be replaced or not
  * @property {Function} [validateFile] - this function will check if the downloaded file is correct (by default, error if file size is 0)
@@ -111,7 +109,6 @@ const saveFiles = async (client, entries, folderPath, options) => {
     subPath: options.subPath,
     fileIdAttributes: options.fileIdAttributes,
     manifest: options.manifest,
-    postProcess: options.postProcess,
     contentType: options.contentType,
     shouldReplaceFile: options.shouldReplaceFile,
     validateFile: options.validateFile || defaultValidateFile,
@@ -330,9 +327,6 @@ const saveFile = async function (client, entry, options) {
     attachFileToEntry(resultEntry, createdFile)
 
     sanitizeEntry(resultEntry)
-    if (options.postProcess) {
-      await options.postProcess(entry)
-    }
   } catch (err) {
     if (err.message === 'MAIN_FOLDER_REMOVED') {
       throw err
@@ -731,7 +725,6 @@ function defaultValidateFile(fileDocument, options) {
 }
 
 function sanitizeEntry(entry) {
-  delete entry.fetchFile
   delete entry.requestOptions
   delete entry.filestream
   delete entry.shouldReplaceFile
--- a/packages/cozy-clisk/src/launcher/saveFiles.js
+++ b/packages/cozy-clisk/src/launcher/saveFiles.js
@@ -107,6 +107,7 @@ const saveFiles = async (client, entries, folderPath, options) => {
         console.log(`${level}: saveFiles: ${label}: ${msg}`)
       }
     },
+    retry: options.retry,
     subPath: options.subPath,
     fileIdAttributes: options.fileIdAttributes,
     manifest: options.manifest,
--- a/packages/cozy-clisk/src/launcher/saveFiles.js
+++ b/packages/cozy-clisk/src/launcher/saveFiles.js
@@ -8,9 +8,13 @@ import { dataUriToArrayBuffer } from '../libs/utils'
  * @typedef saveFilesEntry
  * @property {string} [fileurl] - url where to download the corresponding file
  * @property {string} [filename] - name of the file
+ * @property {string} [dataUri] - base64 representation of the content of the file
  * @property {string | ArrayBuffer} [filestream] - name of the file
  * @property {boolean} [_cozy_file_to_create] - Internal use to count the number of files to download
  * @property {object} [fileAttributes] - metadata attributes to add to the resulting file object
+ * @property {string} [subPath] - subPath of the destination folder path where to put the downloaded file
+ * @property {import('cozy-client/types/types').IOCozyFile} [existingFile] - already existing file corresponding to the entry
+ * @property {boolean} [shouldReplace] - result of the shouldReplaceFile function on the entry
  */
 
 /**
@@ -46,6 +50,9 @@ import { dataUriToArrayBuffer } from '../libs/utils'
  * @property {number} [retry] - number of retries if the download of a file failes. No retry by default
  * @property {string} [folderPath] - path to the destination folder
  * @property {Map<import('cozy-client/types/types').FileDocument>} existingFilesIndex - index of existing files
+ * @property {object} sourceAccountOptions - source account options
+ * @property {object} sourceAccountOptions.sourceAccount - source account object _id
+ * @property {object} sourceAccountOptions.sourceAccountIdentifier - source account unique identifier (mostly user login)
  */
 
 /**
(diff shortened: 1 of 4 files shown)


packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -64,7 +62,7 @@ const saveFiles = async (client, entries, folderPath, options) => {
     throw new Error('Savefiles : No list of files given')
   }
   if (entries.length === 0) {
-    log.warn('No file to download')
+    options.log('warn', 'saveFiles', 'No file to download')
```
Crash--, line 67: do we always have this log from option?

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -386,22 +452,26 @@ const shouldReplaceFile = async function (file, entry, options) {
...
+    options.log('error', 'removeFile', 'No client, impossible to delete file')
   } else {
     await client.collection('io.cozy.files').deleteFilePermanently(file._id)
   }
 }
 
 module.exports = saveFiles
 
-function getFileName(entry) {
+function getFileName(entry, options) {
```
Crash--, line 541: add the JSDoc? 🙏

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -386,22 +452,26 @@ const shouldReplaceFile = async function (file, entry, options) {
   return shouldReplaceFileFn(file, entry, options)
 }
 
-const removeFile = async function (client, file) {
+const removeFile = async function (client, file, options) {
```
Crash--, line 531: add the JSDoc? 🙏

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -411,28 +481,34 @@ function sanitizeFileName(filename) {
   return filename.replace(/^\.+$/, '').replace(/[/?<>\\:*|":]/g, '')
 }
 
-function checkFileSize(fileobject) {
+function checkFileSize(fileobject, options) {
```
Crash--, line 560: add the JSDoc? 🙏

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -411,28 +481,34 @@ function sanitizeFileName(filename) {
...
-    log.warn('BAD_FILE_SIZE')
+    options.log('warn', 'checkFileSize', `${name} is empty`)
+    options.log('warn', 'checkFileSize', 'BAD_FILE_SIZE')
     return false
   }
   return true
 }
 
-function logFileStream(fileStream) {
+function logFileStream(fileStream, options) {
```
Crash--, line 571: add the JSDoc? 🙏

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -465,8 +541,8 @@ function calculateFileKey(entry, fileIdAttributes) {
     .join('####')
 }
 
-function defaultValidateFile(fileDocument) {
-  return checkFileSize(fileDocument)
+function defaultValidateFile(fileDocument, options) {
```
Crash--, line 620: add the JSDoc? 🙏

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -496,20 +572,10 @@ function getFilePath({ file, entry, options }) {
...
-
-async function getOrCreateDestinationPath(client, entry, saveOptions) {
-  const subPath = entry.subPath || saveOptions.subPath
-  let finalPath = saveOptions.folderPath
-  if (subPath) {
-    finalPath += '/' + subPath
-    await client.collection('io.cozy.files').createDirectoryByPath(finalPath)
-  }
-  return finalPath
-}
```
Crash--, line 515: Should be part of this commit? It seems that we use `getOrCreateDestinationPath` that is undefined now?

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -157,6 +161,44 @@ const saveFiles = async (client, entries, folderPath, options) => {
   return savedEntries
 }
 
+/**
+ * Ensure the existence of all destination folders : folderPath, options.subPath and all subPaths
+ * which may be present in each entries.
+ * If the user changes some folders during the execution of saveFile, we will catch and fix the
+ * errors.
+ */
+async function ensureAllDestinationFolders({
```
Crash--, line 171: jsdoc

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -157,6 +161,44 @@ const saveFiles = async (client, entries, folderPath, options) => {
...
+  pathsList.push(...entriesPathList)
+
+  // now ensure paths and put their ids in folderMap
+  for (const path of pathsList) {
+    const createPathResult = await fileCollection.createDirectoryByPath(path)
+    folderMap.set(path, createPathResult.data._id)
+  }
+
+  return folderMap
+}
```
Crash--, line 201: Can't you use https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/FileCollection.js#L865-L871 for instance?

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -301,13 +343,46 @@ function noMetadataDeduplicationErrors(options) {
...
+ * retry createFile. This is done for the case where the destination folder is removed during the
+ * execution of the konnector
+ */
+async function createFileWithFolderOnError(
+  client,
+  entry,
+  options,
+  method,
+  file
+) {
```
Crash--, line 358: jsdoc

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -301,13 +343,46 @@ function noMetadataDeduplicationErrors(options) {
...
+
 async function createFile(client, entry, options, method, file) {
-  const folder = await client
-    .collection('io.cozy.files')
-    .statByPath(options.folderPath)
+  const subPath = entry.subPath || options.subPath
+  const finalPath = subPath
+    ? options.folderPath + '/' + subPath
+    : options.folderPath
+  const dirId = options.folderMap.get(finalPath)
```
Crash--, line 383: We should not rely on that. Can't you pass the `dirId` to the createFile method? Since you have it from here: 

https://github.com/konnectors/libs/pull/973/commits/77234ae36ecb5188b83b97e540bea960148c938e#diff-fa7929926d4c673754905d5b4b3b9f669c0d61759edf0cc75dfd7b0a22e733f3R367-R369

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -301,13 +343,46 @@ function noMetadataDeduplicationErrors(options) {
...
+  client,
+  entry,
+  options,
+  method,
+  file
+) {
+  try {
+    return await createFile(client, entry, options, method, file)
+  } catch (err) {
+    if (err.response?.status === 404) {
```
Crash--, line 362: We just need the path here if there is something wrong.

packages/cozy-clisk/src/launcher/saveFiles.js
doubleface: This saveFiles is called by the flagship app and cozy-clisk is included in the flagship app.Another PR will come with the update of cozy-clisk and `log` option.

Crash--: Yep, but in reality this change is a breaking change. You have to edit the caller code in order to make it work without breaking (aka here throwing an error `options.log is not a function`). 

So tag it as a BC or make it optional by checking if these is an `options.log` before calling it

doubleface: Then made it optionnal

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -125,14 +134,9 @@ const saveFiles = async (client, entries, folderPath, options) => {
...
     }
-    const folderPath = await getOrCreateDestinationPath(
-      client,
-      entry,
-      saveOptions
-    )
     entry = await saveFile(client, entry, {
       ...saveOptions,
-      folderPath
+      folderMap
```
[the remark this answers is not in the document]
doubleface: I pass the folderMap here to be able to modify it in case we have to create the folder on error in `createFileWithFolderOnError`. If we create a folder, we have to update the folderMap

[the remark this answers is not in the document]
Crash--: Can you explain the goal of `folderMap` please? It seems to me that this is an anti-pattern (you edit the folderMap from methods so your methods have side-effect and it's not wanted). Why do we need it? 

If this is to be sure that the `path` is created, you should only use https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/FileCollection.js#L865-L871. So the need is only to have the `finalPath` (aka `folderPath`) when calling `createFileWithFolderOnError` right?

packages/cozy-clisk/src/launcher/saveFiles.js
doubleface: Here we can do it yes. Done

packages/cozy-clisk/src/launcher/saveFiles.js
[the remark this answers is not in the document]
doubleface: This is exactly the point of this PR. I don't want to call `ensureDirectoryExists` or `createDirectoryByPath` for each file because it can take a lot of time (multiplied by the number of files to save).

Since :
- the final `client.save` of the file needs a `dirId` attribute to specify in which folder the file will be saved, 
- each file can have a different destination folder (because of the `subPath` attribute)
- the user could remove a destination folder during the execution of the konnector

I need to : 

- keep an association of paths to their id : the folderMap, which avoids a lot of calls to CozyClient
- update this folderMap if a destination folder has been removed during the konnector execution to have the new dirId associated to the destination folder.

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -262,13 +362,92 @@ function noMetadataDeduplicationErrors(options) {
...
+      }
+    }
+    throw err
+  }
+}
+
+/**
+ *
+ * @param {object} options - options object
+ * @param {import('cozy-client/types/CozyClient')} options.client - CozyClient instance
```
Crash--, line 439: If you import 
```
 * @param {import('cozy-client/types/CozyClient').default} options.client - CozyClient instance
```

you can remove the ts-ignore on .save ;)

packages/cozy-clisk/src/launcher/saveFiles.js
```diff
@@ -386,22 +595,41 @@ const shouldReplaceFile = async function (file, entry, options) {
...
+ * @param {import('cozy-client/types/CozyClient')} client - CozyClient instance
+ * @param {import('cozy-client/types/types').FileDocument} file - file to remove
+ * @param {saveFileOptions} options - options object
+ */
+const removeFile = async function (client, file, options) {
   if (!client) {
-    log.error('No client, impossible to delete file')
+    options.log('error', 'removeFile', 'No client, impossible to delete file')
   } else {
+    // @ts-ignore Property 'collection' does not exist on type 'CozyClient"
```
Crash--, line 609: same here with the right import ;)

Crash-- (approved):
Nice work ! If you import the `.default` for the cozy-client type, then you can remove all the `@ts-ignore`

PR #980, opened 2023-12-12
feat: Add waitForDomReady method to ContentScript

doubleface:
This method will be called when the DOMContentLoaded event is emitted or
if it has already been emitted.

Since this is a pattern we begin to see in every clisk konnectors, let's
avoid to each konnector to implement it :

```javascript
async onWorkerReady() {
  await this.waitForDomReady()
  document.body.addEventListener('click', ...)
}
```

We still add a timeout of 10s to this method, with a warning message (no
error or Exception), to be able to trace any missed event.


KillianCourvoisier (approved):

packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -164,7 +164,27 @@ export default class ContentScript {
...
+            log.warn(
+              'Could not find any body in the current page after 1s',
+              window.location.href
+            )
+          }
+        }
+      })
+      if (found) {
+        this.onWorkerReady()
+      }
```
Crash--, line 187: Can't you use https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#onloadend instead of relying on that? Adding a completely random timeout is not great.

doubleface: From my tests, at least on android, it looks like onLoadEnd is called before the init of the worker, then at this step it is impossible to call the worker to send it a message.

Crash-- (approved):

PR #986, opened 2023-12-26
[SCR-15] Migration node20

LucsT:
This PR will do dual build for node 20 and node 16 engines.

Eslint-plugin-jsdoc has been removed because newer version (45.0.X) are now compatible with node >=18, but exclude node 16.
Some tests certs have been modify to be compatible.
Some more generic node version are use (due to some missing at travis), not a precise one
Build now done in ubuntu focal, due to LibC problems at travis.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/request/request.spec.js
+++ b/packages/cozy-konnector-libs/src/libs/request/request.spec.js
@@ -144,16 +144,28 @@ describe('requestFactory', () => {
       try {
         await rq('https://untrusted-root.badssl.com')
       } catch (err) {
-        expect(err.message).toEqual(
-          'Error: self-signed certificate in certificate chain'
+        /* Second string matching is for node 16 retro compatibility
+           In node 16, error does not contain an hyphen
+        */
+        expect(err.message).toMatch(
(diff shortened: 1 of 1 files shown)


Crash--:
We want to support node 20 & node 16 during the migration path. Please see how we do that for our apps.

packages/cozy-konnector-libs/src/libs/request/request.spec.js
```diff
@@ -144,16 +144,23 @@ describe('requestFactory', () => {
...
+          /Error: self-signed certificate in certificate chain|Error: self signed certificate in certificate chain/
         )
       }
     })
 
     test('Self-signed cert should be refused', async () => {
-      return expect(rq('https://self-signed.badssl.com')).rejects.toEqual(
-        new Error('Error: self-signed certificate')
-      )
+      if (process.version.startsWith('v16')) {
```
doubleface, line 155: I cannot see any difference between the two alternatives

doubleface: Ok, let's add a comment on this to avoid to have someone remove it later because he/she does not see the point.

doubleface (approved):
Let's add a comment on the hyphen thing in tests and it's ok for me

PR #987, opened 2023-12-28
fix: saveIdentity tries to take care of malformed data

Crash--:
If strings are passed when we want array, we try to fix them. We also add warns in order to fix the root cause.

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/saveIdentity.js
+++ b/packages/cozy-konnector-libs/src/libs/saveIdentity.js
@@ -71,14 +71,7 @@ const saveIdentity = async (
     : { contact: contactOrIdentity }
   identity.identifier = accountIdentifier
 
-  // Format contact if needed
-  if (identity.contact.phone) {
-    identity.contact.phone = formatPhone(identity.contact.phone)
-  }
-  if (identity.contact.address) {
-    identity.contact.address = formatAddress(identity.contact.address)
-  }
-
+  identity.contact = formatIdentityContact(identity.contact)
   if (options.merge == false) {
     // Using cozy client in a none merging strategy here.
     const newClient = cozyClient.new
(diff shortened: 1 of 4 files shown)


packages/cozy-konnector-libs/src/libs/saveIdentity.js
```diff
@@ -145,9 +140,57 @@ function formatPhone(phone) {
...
-module.exports = saveIdentity
+function formatIdentityContact(contactToFormat) {
+  const contact = { ...contactToFormat }
+  if (contact.phone) {
+    if (!Array.isArray(contact.phone)) {
+      log(
+        'warn',
+        'formatIdentityContact Phone is not an array, you should fix it'
+      )
+      contact.phone = [
```
KillianCourvoisier, line 158: If I may, here we usually use the following structure :

```
const phone = [
  {
   number : 0000000000,
   type : 'mobile'
  }
]
```
The type is determine with the first two numbers of the phoneNumber.
Maybe we could had it here too, to prevent further issues ?

Something like : 

```
if (contact.phone) {
    if (!Array.isArray(contact.phone)) {
      log(
        'warn',
        'formatIdentityContact Phone is not an array, you should fix it'
      )
      let type
      if(contact.phone.startsWith('06') || contact.phone.startsWith('07'){
        type = 'mobile
      }else{
        type = 'home'
      }
      contact.phone = [
      {
          number: contact.phone,
          type
        }
      ]
    }
    contact.phone = formatPhone(contact.phone)
  
```
What do you think ?

Crash--: By reading https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.contacts.md#iocozycontacts 

`type` is not mandatory. So I don't want to add black magic here. You can have a `work` mobile phone. This is something to handle at the konnector level. Here we don't know anything.

KillianCourvoisier: Alright then

KillianCourvoisier (approved):

PR #990, opened 2024-01-18
[SCR-99] fix: Fix force qualification update

LucsT:
Fix repeated downoad of files with cozy-konnector-libs

### DIFF
--- a/packages/cozy-konnector-libs/src/libs/saveFiles.js
+++ b/packages/cozy-konnector-libs/src/libs/saveFiles.js
@@ -10,11 +10,14 @@ const path = require('path')
 const requestFactory = require('./request')
 const omit = require('lodash/omit')
 const get = require('lodash/get')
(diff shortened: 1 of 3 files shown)


packages/cozy-konnector-libs/src/libs/saveFiles.js
```diff
@@ -502,9 +505,15 @@ const shouldReplaceFile = async function (file, entry, options) {
   }
   const defaultShouldReplaceFile = (file, entry) => {
     const shouldForceMetadataAttr = attr => {
-      const result =
-        getAttribute(file, `metadata.${attr}`) !==
-        get(entry, `fileAttributes.metadata.${attr}`)
+      let entryQualif = get(entry, `fileAttributes.metadata.${attr}`)
```
Crash--, line 508: this is not `entryQualif`, this is `entryMetadataAttribute` . Non ?

PR #991, opened 2024-01-19
[SCR-138] feat: Add RequestInterceptor

doubleface:
Allows the konnectors to wait for any http request interception and more + a documentation

### DIFF
--- a/packages/cozy-clisk/docs/RequestInterceptor.md
+++ b/packages/cozy-clisk/docs/RequestInterceptor.md
@@ -1,0 +1,77 @@
+# RequestInterceptor
+
+## Intercepting HTTP requests in clisk konnectors
+
+To intercept HTTP requests, you can init the RequestInterceptor like this :
+
+```javascript
+import {
+  ContentScript,
+  RequestInterceptor
+} from 'cozy-clisk/dist/contentscript'
+
+const requestInterceptor = new RequestInterceptor([
+  {
+    label: 'testRequest',
+    method: 'GET',
+    url: 'test/api',
+    serialization: 'json'
+  }
+])
+requestInterceptor.init()
+
+const connector = new ContentScript({ requestInterceptor })
+```
+
+This will intercept all requests with method 'GET' and containing 'test/api' in their url. The
+response  of these requests will be serialized to a json object.
+
+## Fetching the response
+
+You can fetch the intercepted responses in two ways : `waitForRequestInterception` or directly
+listening to requestInterceptor events.
+
+
+### waitForRequestInterception
+
+Once you have defined an interception with a given `testRequest` label, you can wait for it to
+happen using the following from the pilot:
+
+```javascript
+const testRequestResponse = await this.waitForRequestInterception('testRequest')
+```
+
+It is important that waitForRequestInterception is run before the real request response to be
+received. Or else you won't get anything and the promise will reject with a Timeout error.
+
+`testRequestResponse` will have the form :
+
+```javascript
+{
+  label: 'testRequest',
+  url: '...', // the real intercepted url
+  method: 'GET',
+  response : {}, // a javascript object in our case
+  responseHeaders : {} // a javascript object with all response headers
+  requestHeaders : {} // a javascript object with all request headers
+}
+```
+
+### RequestInterceptor events
+
+Each the the RequestInterceptor object intercepts an expected request, defined in it's
+instanciation, it will emit an event that you can intercept in onWorkerEvent method from the pilot
+ like this:
+
+ ```javascript
+ onWorkerEvent({ event, payload }) {
+  if (event === 'requestResponse') {
+    // the payload contains the whole response like defined in the previous section
+  }
+}
+ ```
+
+
+
+
+
--- a/packages/cozy-clisk/src/contentscript/ContentScript.js
+++ b/packages/cozy-clisk/src/contentscript/ContentScript.js
@@ -33,7 +33,7 @@ if (window?.addEventListener) {
 }
 
 export default class ContentScript {
-  constructor() {
+  constructor(options = {}) {
     sendPageMessage('NEW_WORKER_INITIALIZING')
     const logDebug = message => this.log('debug', message)
     const wrapTimerDebug = wrapTimerFactory({ logFn: logDebug })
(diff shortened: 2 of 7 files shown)


KillianCourvoisier (approved):

packages/cozy-clisk/docs/RequestInterceptor.md
```diff
@@ -0,0 +1,77 @@
...
+    url: 'test/api',
+    serialization: 'json'
+  }
+])
+requestInterceptor.init()
+
+const connector = new ContentScript({ requestInterceptor })
+```
+
+This will intercept all requests with method 'GET' and containing 'test/api' in their url. The
```
Crash--, line 26: so this is `*test/api*` right?

packages/cozy-clisk/docs/RequestInterceptor.md
```diff
@@ -0,0 +1,77 @@
...
+response  of these requests will be serialized to a json object.
+
+## Fetching the response
+
+You can fetch the intercepted responses in two ways : `waitForRequestInterception` or directly
```
Crash--, line 31: response

packages/cozy-clisk/docs/RequestInterceptor.md
```diff
@@ -0,0 +1,77 @@
...
+  ContentScript,
+  RequestInterceptor
+} from 'cozy-clisk/dist/contentscript'
+
+const requestInterceptor = new RequestInterceptor([
+  {
+    label: 'testRequest',
+    method: 'GET',
+    url: 'test/api',
+    serialization: 'json'
```
Crash--, line 18: What is it? Is it the `content/type` ? the accepted header?

packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -165,8 +182,22 @@ export default class ContentScript {
...
 
+    if (!this.bridge) {
+      throw new Error(
+        'No bridge is defined, you should call ContentScript.init before using this method'
+      )
+    }
+
     if (contentScriptType === WORKER_TYPE) {
+      this.requestInterceptor?.on('response', response => {
+        this.bridge?.emit('workerEvent', {
```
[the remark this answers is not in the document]
Crash--: Isn't weird to add this listener before calling workerready ?

packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -268,6 +299,44 @@ export default class ContentScript {
...
+    const timeout = options?.timeout ?? 60000
+
+    const interceptionPromise = new Promise(resolve => {
+      const listener = ({ event, payload }) => {
+        if (event === 'requestResponse' && payload.label === label) {
+          if (!this.bridge) {
+            throw new Error(
+              'No bridge is defined, you should call ContentScript.init before using this method'
+            )
+          }
```
Crash--, line 321: no need to add this test, you already added it before adding the listener

packages/cozy-clisk/src/contentscript/ContentScript.spec.js
```diff
@@ -5,11 +5,11 @@ describe('ContentScript', () => {
   describe('saveFiles', () => {
     it('should call launcher saveFiles with given nominal options', async () => {
       const contentScript = new ContentScript()
-      contentScript.setContentScriptType(PILOT_TYPE)
       contentScript.bridge = {
         call: jest.fn(),
         emit: jest.fn()
       }
+      contentScript.contentScriptType = PILOT_TYPE
```
Crash--, line 12: why this change?

packages/cozy-clisk/docs/RequestInterceptor.md
doubleface: I chose to make these stars implicit, you think we should accept glob patterns or regexp ?

packages/cozy-clisk/docs/RequestInterceptor.md
```diff
@@ -0,0 +1,77 @@
...
+
+```javascript
+import {
+  ContentScript,
+  RequestInterceptor
+} from 'cozy-clisk/dist/contentscript'
+
+const requestInterceptor = new RequestInterceptor([
+  {
+    label: 'testRequest',
```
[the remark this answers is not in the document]
doubleface: Right, I think identifier will be better

packages/cozy-clisk/docs/RequestInterceptor.md
doubleface: This is how we choose to serialize the body of the response, since it needs to be serialized to be transmitted from the worker to the pilot. We could relate it to the content type but I preferred it to be explicit.

packages/cozy-clisk/src/contentscript/ContentScript.js
doubleface: This was actually ts-check which enforces me to do it. In theory, nothing avoid to remove the bridge between the two tests :shrug:

packages/cozy-clisk/src/contentscript/ContentScript.spec.js
doubleface: To avoid the call to setContentScript in this test, since setContentScript does more now.

packages/cozy-clisk/src/contentscript/ContentScript.js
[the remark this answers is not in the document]
doubleface: 1) typechecking too
2) right ! The konnector should be able to subscribe to these event before on event is emitted. I will fix it

packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -165,8 +182,22 @@ export default class ContentScript {
...
+    if (!this.bridge) {
+      throw new Error(
+        'No bridge is defined, you should call ContentScript.init before using this method'
+      )
+    }
+
     if (contentScriptType === WORKER_TYPE) {
+      this.requestInterceptor?.on('response', response => {
+        this.bridge?.emit('workerEvent', {
+          event: 'requestResponse',
```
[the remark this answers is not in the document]
doubleface: Since these are events already coming from RequestInterceptor, I found it redundant, no ?

packages/cozy-clisk/docs/RequestInterceptor.md
Crash--: I don't like the implicit, and I think we should avoid it. I don't know what we should accept, but regex seems a good candidate.

packages/cozy-clisk/src/contentscript/ContentScript.js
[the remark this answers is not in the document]
Crash--: When you have a large codebase and you start looking for something, I can assume you that there is nothing redundant. Because looking for `requestResponse` will get you a lot of responses. And the majority will not be related to the ContentScript. So you'll have to look for `requestInterceptor` and then you'll find a few places where it's used. And then you'll have to read those places to look for `requestResponse`. 

I know that is this the idea of the grep test (https://jamie-wong.com/2013/07/12/grep-test/) taken to the extreme, but this is how I can navigate into the all cozy's codebase.

packages/cozy-clisk/src/contentscript/ContentScript.spec.js
Crash--: What else does it do?

PR #992, opened 2024-01-22
chore: use cozy-client as peerDep in cozy-clisk (VO-265)

acezard:
Not doing so can introduce breaking behaviour in the consuming app if it's using a different version of cozy-client

### DIFF
--- a/packages/cozy-clisk/package.json
+++ b/packages/cozy-clisk/package.json
@@ -36,11 +36,13 @@
   "dependencies": {
     "@cozy/minilog": "^1.0.0",
     "bluebird-retry": "^0.11.0",
-    "cozy-client": "^41.2.0",
     "ky": "^0.25.1",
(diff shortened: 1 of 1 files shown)


Crash--:
This is a fix. Not a chore. With a core you don't publish a new version since you should don't change the built code. Here you want a new version, so this is not a chore ;). 

To the  @konnectors/cozy team, cozy-* should never be set as a dep.  cf https://github.com/cozy/cozy-guidelines/#dependencies

doubleface (approved):

acezard:
> This is a fix. Not a chore. With a core you don't publish a new version since you should don't change the built code. Here you want a new version, so this is not a chore ;).
> 
> To the @konnectors/cozy team, cozy-* should never be set as a dep. cf https://github.com/cozy/cozy-guidelines/#dependencies

Updated 👍

PR #994, opened 2024-01-25
fix: RequestInterceptor enhancements after review

doubleface:
- refactor: Rename interception.label to interception.identifier
- docs: enhancements
- fix: Send request interception events after onWorkerReady


Crash-- (approved):
Thanks <3

PR #997, opened 2024-02-16
[SCR-303] feat(cozy-konnector-libs): Add more verbose option for saveFiles results

LucsT:
Adding the possibility for a connector to pass an option that will convert the log result of saveFiles from a info level to a warning level.

Useful for studying download consistency. on a connecteur at big scale.


doubleface (approved):

doubleface:
Ok for me when the lint is ok

PR #999, opened 2024-02-21
[SCR-331] fix: replace 'label' by 'identifier' for requestInterception

KillianCourvoisier:
Fixes an oversight regarding option naming for request interceptions. This was forgotten and seems to lead konnectors which uses the `waitForRequestInterception` to crash on never resolving function.

### DIFF
--- a/packages/cozy-clisk/src/contentscript/ContentScript.js
+++ b/packages/cozy-clisk/src/contentscript/ContentScript.js
@@ -300,20 +300,20 @@ export default class ContentScript {
   }
 
   /**
-   * Wait for the given labelled request to be intercepted. The labeled request must be defined and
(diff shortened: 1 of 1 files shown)


packages/cozy-clisk/src/contentscript/ContentScript.js
```diff
@@ -303,17 +303,17 @@ export default class ContentScript {
    * Wait for the given labelled request to be intercepted. The labeled request must be defined and
```
doubleface, line 303: Replace labelled with identified

KillianCourvoisier: Oopsi, indeed.

There's is something else with label, the build won't work, i'm looking for it 👍

doubleface (approved):

